Issue #5, opened 2018-02-02
Enable CI on licensed


jonabc:
Enable CI to run tests to validate pull requests.  Travis CI and Circle CI look like the best options at the moment.

This will be require some extra thought and work due to the nature of the project.  For each supported package manager, language or tool that's used we need to ensure that it is available on in the CI environment so that tests can run.

There's also the question of having one monolithic CI job that installs all dependencies runs all source tests, or breaking out each source into it's own CI job.

My 2c is to try and modularize this as much as possible and use separate CI jobs for each source.

PR #12, opened 2018-03-23
Separate test fixture setup scripting from bootstrap

jonabc:
This PR moves the test fixture setup for each source to separate scripts as part of [enabling CI](https://github.com/github/licensed/issues/5).  This makes it easier to define CI jobs for each source that test in isolation.

This is mostly a subset of the changes in https://github.com/github/licensed/commits/ci-travis where Travis CI was enabled and 💚 

I went back and forth on whether to separate the scripts or continue to use a single `bootstrap` script to set everything up when the needed tools were present.  I decided on this approach because it makes the following easy
1. running individual source setups is straightforward and not manipulated by env vars
2. setup fails in CI if required environment is not available
3. a single entry point can run all setup scripts (in this case `rake setup`)
   - "command not found" failures can be caught and treated as warnings

Also fixed some errors in the go test fixture surfaced by `go get` that were previously ignored, and some warnings from the npm test fixture that were chatty.  Running `rake setup` on a machine with all tools available shouldn't produce any warnings or errors.

/cc @benbalter you left some feedback on https://github.com/github/licensed/issues/5 so I'm curious for your 💭 here 🙇


CONTRIBUTING.md
```diff
@@ -13,7 +13,8 @@ Please note that this project is released with a [Contributor Code of Conduct][c
 
 0. [Fork][fork] and clone the repository
 0. Configure and install the dependencies: `script/bootstrap`
-0. Make sure the tests pass on your machine: `rake`
+0. Setup test fixtures: `rake setup`
+0. Make sure the tests pass on your machine: `rake test`
```
benbalter: To be consistent with ["scripts to rule them all" expectations](https://github.com/github/scripts-to-rule-them-all), I'd suggest creating a quick `script/test` file, even if all it does it run `bundle exec rake test`.

script/setup/bower
```diff
@@ -0,0 +1,12 @@
...
+
+if [ -z "$(which bower)" ]; then
+  echo "A local bower installation is required for bower development." >&2
+  exit 127
+fi
+
+# setup test fixtures
+BASE_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+cd $BASE_PATH/test/fixtures/bower
+bower install
```
benbalter: It's semantics, but I'd expect `script/bootstrap` to install all dependencies with `script/setup` creating fixtures, getting the DB into the proper state, etc.

> script/bootstrap is used solely for fulfilling dependencies of the project.
>
> This can mean RubyGems, npm packages, Homebrew packages, Ruby versions, Git submodules, etc.
>
> The goal is to make sure all required dependencies are installed.

Specifically, would it make sense to rename the `script/setup/*` individual files to `script/bootstrap/*` files, and call `rake setup` (or `rake bootstrap`) from `script/bootstrap?

jonabc: > It's semantics, but I'd expect script/bootstrap to install all dependencies with script/setup creating fixtures

These setup scripts are creating fixtures ☝️ 

I'd expect bootstrap scripts to be installing tools that are required (cabal/ghc, go, etc) for licensed development, which is outside what the setup scripts are doing.

benbalter: :+1: That makes total sense. Thanks for talking through it with me.

benbalter: Similar to the above, I might add a quick `script/setup` script as well, (even if it just runs `rake setup`).

jonabc: @benbalter thanks for the feedback.  I had to rename the existing `script/setup` folder to not clash with the new script, and settled on `script/source-setup`.  Do you know of any naming precedence that might apply here?

I also thought about naming all the scripts like `script/setup-<source>` but 🤷‍♂️ I liked the folder better for nesting

PR #13, opened 2018-03-23
Command source-agnostic tests and integration tests

jonabc:
As part of [enabling CI](https://github.com/github/licensed/issues/5), [test fixture setups are moving into separate scripts](https://github.com/github/licensed/pull/12) so that CI jobs can test dependency sources in isolation.

Along with ☝️, some changes are needed for the source-agnostic `command` tests.  When the source setups are run independent of `script/bootstrap`, we can't hardcode the command tests to a single real dependency source because it may not be setup for a given CI job.

This PR solves that by making most of the command tests agnostic of any real dependency sources by using a mock `TestSource`.

A test for each of the `cache` and `list` commands has been made dependent to run over every available dependency source.  The available source types can be filtered via an ENV var so that only the requested dependency source is tested.  This means that during CI all tests are either source agnostic, or use the ENV-specified dependency source for the given CI job.

**Note** There are some issues that need to be fixed with the bundler and go sources before they can be included in these integration tests, I've made TODO notes of them [here](https://github.com/github/licensed/compare/test-source?expand=1#diff-f14ca9959ff06f3ca2ee6b9564866954R51) and will be opening issues for the work.  I don't think that should block this PR and the overall goal of enabling CI.

Once this and https://github.com/github/licensed/pull/12 are 👍 , setting up the travis configuration should be fairly straightforward.

/cc @benbalter FYI.  If you have a few moments to sanity check this I would appreciate it!


test/command/cache_test.rb
```diff
@@ -3,44 +3,67 @@
...
-    license = Licensed::License.read(config.cache_path.join("rubygem/licensee.txt"))
-    assert_equal "licensee", license["name"]
-    assert_equal "mit", license["license"]
+  after do
+    config.apps.each do |app|
+      FileUtils.rm_rf app.cache_path
+    end
+  end
+
+  each_source do |source_type|
```
jonabc: the configuration and expectations for each dependency source is loaded from the new yaml fixtures.  This seemed like the easiest and cleanest way to make these tests dynamic to the requirements for each dependency source

test/command/list_test.rb
```diff
@@ -3,23 +3,41 @@
...
-    config.sources.each do |s|
-      assert_match(/#{s.type} dependencies:/, out)
+  before do
+    config.apps.each do |app|
+      app.sources.clear
+      app.sources << source
+    end
+  end
+
+  each_source do |source_type|
```
jonabc: same as above here - using a configuration fixture seemed like the best approach here

test/test_helper.rb
```diff
@@ -18,3 +18,43 @@
     Licensed.use_github = false
   end
 end
+
+class TestSource
```
jonabc: This mock was made as lightweight and simple as possible.  It could probably be beefed up a bit and made more flexible but we can leave that to when it's needed 👍

jonabc:
@benbalter I'm planning to merge this today unless you would like me to hold off so you have a chance to take a look.  I'd like to get CI enabled before moving forward with any feature work.

The issues with the `go` and `bundler` sources that were noted above have been fixed.

PR #19, opened 2018-03-29
Enable Travis CI

jonabc:
Resolves https://github.com/github/licensed/issues/5

This turns on Travis CI and runs a job for each dependency source.  Each job runs

1.  Tests specific to that source
2. Tests for the non-source licensed components

This strategy errs on the side of caution to make sure that dependency sources don't modify and break source-agnostic code.  Structuring the tests this way also makes development easier on a single source - running `bundle exec rake test:[source]` will run a full test pass on `licensed` for the source.

This PR also makes the following changes that were needed to get CI working

1. test suites can be run from either rake tasks or `script/test`

```
# the following are equivalent
bundle exec rake test:go
./script/test go
```

2. fixed a few issues that were causing problems in CI
   1. https://github.com/github/licensed/commit/c366e24f51be2448c0ecee3db211ba5017b40abf forces CLI tests to run offline and avoid timeouts from querying GitHub in CI.
   2. https://github.com/github/licensed/commit/d858334833745786bd64623fe858fa0463573245 fixes a problem where a nil result from querying GitHub was not cached properly.  This was an initial fix attempt for ☝️ 
   3. https://github.com/github/licensed/commit/83f572f0f56c701c5f3a22f32f70432c8b6ba72f clears `BUNDLE_GEMFILE` from ENV in the `bundler` source.  The variable is set during CI and was causing the wrong Gemfile to be used when enumerating dependencies

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 require "bundler/gem_tasks"
 require "rake/testtask"
+require "rubocop/rake_task"
 
 desc "Run source setup scripts"
 task :setup do
@@ -51,6 +52,12 @@ Rake::TestTask.new(:test) do |t|
   t.libs << "test"
   t.libs << "lib"
   t.test_files = FileList["test/**/*_test.rb"]
+end
+
+# add rubocop task
+# -S adds styleguide urls to offense messages
+RuboCop::RakeTask.new do |t|
+  t.options.push "-S"
 end
 
 task default: :test
--- a/Rakefile
+++ b/Rakefile
@@ -23,6 +23,30 @@ task :setup do
   end
 end
 
+sources = Dir["lib/licensed/source/*.rb"].map { |f| File.basename(f, ".*") }
+
+namespace :test do
+  sources.each do |source|
+    # hidden task to set ENV and filter tests to the given source
+    # see `each_source` in test/test_helper.rb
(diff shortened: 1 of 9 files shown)


Rakefile
```diff
@@ -23,10 +24,40 @@ task :setup do
   end
 end
 
+sources = Dir["lib/licensed/source/*.rb"].map { |f| File.basename(f, ".*") }
```
benbalter: I _think_ this may be relative to the `pwd`, not the file. To make things a bit more solid, it may be worthwhile to make the path absolute (e.g., `File.expand_path .... __dir__`)

Rakefile
```diff
@@ -23,10 +24,40 @@ task :setup do
...
+  sources.each do |source|
+    # hidden task to set ENV and filter tests to the given source
+    # see `each_source` in test/test_helper.rb
+    namespace source.to_sym do
+      task :env do
+        ENV["SOURCE"] = source
+      end
+    end
+
+    Rake::TestTask.new(source => "test:#{source}:env") do |t|
```
benbalter: Nice.

Rakefile
jonabc: Testing locally, it looks `pwd` is the Gemfile dir when run with `bundler exec` but that ✨ being explicit

Issue #2, opened 2018-02-02
Avoid license redetection when license text hasn't changed


jonabc:
There have been cases where a dependency is re-cached with no changes to the license text, however the license field is changed to `other`.  

If the license text hasn't changed, we should keep the existing license to reduce noise to users.

[`project.license.content_normalized`](https://github.com/benbalter/licensee/blob/master/lib/licensee/content_helper.rb#L67-L72) will return a normalized version of the license file's content. This should theoretically be the "cleanest" comparison, in that ignores non-legally significant differences like whitespace, capitalization, etc.

☝️ can be used to compare the cached license content to the detected license text and retain the existing license if a match is found

Issue #20, opened 2018-04-03
`licensed status` can false-succeed with missing license text


jonabc:
The status command checks `license.text.strip.empty?` when determining whether a cached license metadata file is missing license content.  This can be a false-success in the following scenario

1.  During `licensed cache`, a dependency has no license text but does have legal notices
2. During `licensed cache`, the dependency☝️ is saved with legal notices in `license.text`
3. During `licensed status`, the cached license has non-empty `license.text` (notice file contents).  This results in a pass for the cached metadata even though license text is not in the cached metadata.

PR #21, opened 2018-04-03
Avoid license key change when dependency license text hasn't changed

jonabc:
Resolves https://github.com/github/licensed/issues/2.

This PR affects the `cache` command, attempting to reuse the `license` metadata value when possible to cut down on noise.

`licensed cache` now compares the license available in a dependency's cached file to the content found from `dependency.detect_license!`.  When no changes are found, `licensed cache` will reuse the cached license value.  `License#license_text_match?` uses `Licensee::ContentHelper` to compare normalized license content instead of the raw data.

A side effect of this PR is a slight behavior change that should be entirely backwards compatible - a licensed text section will always be written to the cached file.  We used to strip out `nil` license text and legal notices which means that [`licensed` couldn't know if the first (or only) text in a cached file is the license text or a legal notice](https://github.com/github/licensed/issues/20).

Ensuring that `Dependency#detect_license!` always includes section for license text in `self.text` means that we can reliably use that section to compare cached license text against the license text found from `Dependency#detect_license!`.

This makes the assumption that it is 👍 to only compare license text to determine that the license key can be reused.  If `Dependency#notices` should also be compared, then https://github.com/github/licensed/tree/less-license-chattiness should be considered instead of this PR.

/cc @lildude FYI as `github/linguist` was asking for this.  Please let me know if you'd like me to ping someone else for `linguist`s attention 🙇

### DIFF
--- a/lib/licensed/license.rb
+++ b/lib/licensed/license.rb
@@ -2,10 +2,15 @@
 require "yaml"
 require "fileutils"
 require "forwardable"
+require "licensee"
 
 module Licensed
   class License
+    include Licensee::ContentHelper
+    extend Forwardable
+
     YAML_FRONTMATTER_PATTERN = /\A---\s*\n(.*?\n?)^---\s*$\n?(.*)\z/m
+    TEXT_SEPARATOR = ("-" * 80).freeze
 
     # Read an existing license file
     #
@@ -17,7 +22,6 @@ module Licensed
       new(YAML.load(match[1]), match[2])
     end
 
-    extend Forwardable
     def_delegators :@metadata, :[], :[]=
 
     # The license text and other legal notices
@@ -37,6 +41,27 @@ module Licensed
     def save(filename)
       FileUtils.mkdir_p(File.dirname(filename))
       File.write(filename, YAML.dump(@metadata) + "---\n#{text}")
+    end
+
+    # Returns the raw content for Licensee::ContentHelper#content_normalized.
+    # The content is the license text only without any notices
+    def content
+      return unless text
+
+      # if the text contains the separator, the first string in the array
+      # should always be the license text whether empty or not.
+      # if the text didn't contain the separator, the text itself is the entirety
+      # of the license text
+      split = text.split(TEXT_SEPARATOR)
+      split.length > 1 ? split.first.strip : text
+    end
+
+
+    # Returns whether the current license should be updated to `other`
+    # based on whether the normalized license content matches
+    def license_text_match?(other)
+      return false unless other.is_a?(License)
+      self.content_normalized == other.content_normalized
     end
   end
 end
--- a/lib/licensed/license.rb
+++ b/lib/licensed/license.rb
@@ -18,6 +18,7 @@ module Licensed
     #
     # Returns a Licensed::License
     def self.read(filename)
+      return unless File.exist?(filename)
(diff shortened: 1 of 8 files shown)


mlinksva:
> This makes the assumption that it is 👍 to only compare license text to determine that the license key can be reused. If Dependency#notices should also be compared, then https://github.com/github/licensed/tree/less-license-chattiness should be considered instead of this PR.

Changed notices don't have any behavior impact when license text can be detected, so I don't think they should for this case either, so long as any changed notices actually get updated in the cache.

It might be useful for licensed to report on changed notices or optionally require review of same, but that's a different feature.

lib/licensed/license.rb
```diff
@@ -39,5 +43,26 @@ def save(filename)
...
+      split = text.split(TEXT_SEPARATOR)
+      split.length > 1 ? split.first.strip : text
+    end
+
+
+    # Returns whether the current license should be updated to `other`
+    # based on whether the normalized license content matches
+    def license_text_match?(other)
+      return false unless other.is_a?(License)
+      self.content_normalized == other.content_normalized
```
benbalter: 👍

lildude:
> /cc @lildude FYI as github/linguist was asking for this. Please let me know if you'd like me to ping someone else for linguists attention 🙇

I'm _it_ so this is fiiiiine.

mlinksva:
I'd rather not output "License text is not available" -- that may not be the case -- licensed just may not have found it. If it's meant to be an informative message, maybe something like "Licensed did not find license text; see <link?> for information on where licensed looks."

jonabc:
@mlinksva gotcha.  I used "available" because `Dependency#license_text` is nil when multiple license files are found, so "not found" didn't seem fully appropriate either.

Linking to documentation is helpful but makes the message dependent on an outside resource that could change and mess with content matching.

I'm not sure this is worth pursuing TBH - missing license text is an error case for `licensed status` and users shouldn't allow cached data with empty text.

It was just a thought I had and since the PR deals with what to cache when license text isn't found, it seemed like the right time to bring up.

mlinksva:
> `Dependency#license_text` is nil when multiple license files are found

Hmm, it's a different issue, but I suspect it would be helpful to cache all of the license texts in this case rather than none of them.

jonabc:
> Hmm, it's a different issue, but I suspect it would be helpful to cache all of the license texts in this case rather than none of them.

Agreed, I looked into that but wasn't sure how it would affect, or if it would cause a weird mismatch scenario, with `Dependency#license_key` a.k.a. the `license` metadata value.  I'll open up an issue for that.

jonabc:
>> This makes the assumption that it is 👍 to only compare license text to determine that the license key can be reused. If Dependency#notices should also be compared, then https://github.com/github/licensed/tree/less-license-chattiness should be considered instead of this PR.

> Changed notices don't have any behavior impact when license text can be detected, so I don't think they should for this case either, so long as any changed notices actually get updated in the cache.

@mlinksva sorry I only noticed this now.  What is the expected or required behavior when license text can't be detected?

mlinksva:
@jonabc I was saying the behavior of the code in this PR's branch is fine. Notices should be cached (and updated if changed) for any license state, whether same or changed, detected or not, but notice changes shouldn't invalidate reuse of license key.

Issue #24, opened 2018-04-04
[Bundler] `::Bundler.default_gemfile` finds the wrong Gemfile


jonabc:
`::Bundler.default_gemfile.basename` is used in the [bundler source](https://github.com/github/licensed/blob/9b64114f3880f3e82a79f09b20b26251fa043dae/lib/licensed/source/bundler.rb#L45-L52) when looking for a Gemfile in the applications source path.

The issue is that when licensed is run via `bundle exec`, Bundler sets the `ENV["BUNDLE_GEMFILE"]` variable which is given preference over searching for a Gemfile from the file system.  

This fails when
- the application source path where dependencies are enumerated has a `gems.rb` file
- `bundle exec` runs licensed based on a `Gemfile` in a different location

In this scenario `Bundler#gemfile_path` will return `$pwd/Gemfile` which doesn't exist, rather than `$pwd/gems.rb` which does.

This seems like a low priority edge case, but might be fixable in combination with https://github.com/github/licensed/issues/23?

Issue #23, opened 2018-04-04
[Bundler] `::Bundler.default_gemfile` raising GemFileNotFound error


jonabc:
`::Bundler.default_gemfile.basename` is used in the [bundler source](https://github.com/github/licensed/blob/9b64114f3880f3e82a79f09b20b26251fa043dae/lib/licensed/source/bundler.rb#L45-L52) when looking for a Gemfile in the applications source path.

Using [::Bundler.default_gemfile](https://github.com/bundler/bundler/blob/b9c12ee67661581f8a4fb5f47a375b5c7da2ed1e/lib/bundler/shared_helpers.rb#L32-L36) and [::Bundler.default_lockfile](https://github.com/bundler/bundler/blob/b9c12ee67661581f8a4fb5f47a375b5c7da2ed1e/lib/bundler/shared_helpers.rb#L38-L45) in licensed will raise errors when
1. licensed is not run with an empty or nil `ENV["BUNDLE_GEMFILE"]`
   - e.g. not run from `bundle exec`
```
$> gem install licensed
$> licensed cache ...
```
2. a Gemfile is not available in the file hierarchy from root to the application source path

When both of these are true, a default Gemfile isn't found and `::Bundler.default_gemfile` [raises an error](https://github.com/bundler/bundler/blob/b9c12ee67661581f8a4fb5f47a375b5c7da2ed1e/lib/bundler/shared_helpers.rb#L34).

This can be worked around by explicitly disabling the Bundler source
```
sources:
  rubygem: false
```

PR #27, opened 2018-04-05
[Bundler] Don't use Bundler.default_*file to find gem file and lock file

jonabc:
Fixes https://github.com/github/licensed/issues/23
Fixes https://github.com/github/licensed/issues/24

Fixing a couple issues where `::Bundler.default_*file` does not behave ideally for usage in licensed
1. They raise errors if a gem file can't be found
   - e.g. if the licensed gem is installed globally and not run from bundler, and a gem file isn't in the configured app source path
2. They prefer `ENV["BUNDLE_GEMFILE"]` over searching for a gem file in the local file hierarchy
   - can lead to a mismatch if the name of the gem file in `ENV["BUNDLE_GEMFILE"]` doesn't match the name of the file in the configured app source path (gems.rb vs Gemfile)
3.  licensed has a very specific usage that looks for the gem file at the app source path and does not need any of the extra search logic in [`Bundler::SharedHelpers::find_gemfile`](https://github.com/bundler/bundler/blob/257fb54da0003c3a67f6e7b3b5a242a4ba9c45cb/lib/bundler/shared_helpers.rb#L240-L246)

The downside is that now the source has to either keep track of the possible gem file names itself or call the private [`Bundler::SharedHelpers::gemfile_names`](https://github.com/bundler/bundler/blob/257fb54da0003c3a67f6e7b3b5a242a4ba9c45cb/lib/bundler/shared_helpers.rb#L248-L250).  This PR opts to keep track of the file names locally so that licensed doesn't take a dependency on a private method that could change without notice.

### DIFF
--- a/lib/licensed/source/bundler.rb
+++ b/lib/licensed/source/bundler.rb
@@ -61,9 +61,8 @@ module Licensed
       def with_local_configuration
         # with a clean, original environment
         ::Bundler.with_original_env do
-          # bundler restores all ENV at the end of the `with_original_env`
-          # block.  we shouldn't need to restore these values manually
-          BUNDLER_ENV_KEYS.each { |k| ENV.delete(k) }
+          # force bundler to use the local gem file
+          ENV["BUNDLE_GEMFILE"] = gemfile_path.to_s
 
           # reset all bundler configuration
           ::Bundler.reset!
--- a/lib/licensed/source/bundler.rb
+++ b/lib/licensed/source/bundler.rb
@@ -5,12 +5,14 @@ module Licensed
   module Source
     class Bundler
       BUNDLER_ENV_KEYS = %w(BUNDLE_GEMFILE).freeze
+      GEMFILES = %w{Gemfile gems.rb}.freeze
+
       def initialize(config)
         @config = config
       end
 
       def enabled?
-        @config.enabled?(type) && File.exist?(lockfile_path)
+        @config.enabled?(type) && lockfile_path && lockfile_path.exist?
       end
 
       def type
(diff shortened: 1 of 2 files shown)


lib/licensed/source/bundler.rb
```diff
@@ -57,9 +60,8 @@ def lockfile_path
       def with_local_configuration
         # with a clean, original environment
         ::Bundler.with_original_env do
-          # bundler restores all ENV at the end of the `with_original_env`
-          # block.  we shouldn't need to restore these values manually
-          BUNDLER_ENV_KEYS.each { |k| ENV.delete(k) }
+          # force bundler to use the local gem file
+          ENV["BUNDLE_GEMFILE"] = gemfile_path.to_s
```
dbussink: @jonabc Should this `ENV` var be unset in the `ensure` block below? Otherwise this method that takes a block leaks that environment variable?

test/source/bundler_test.rb
```diff
@@ -1,13 +1,22 @@
 # frozen_string_literal: true
 require "test_helper"
 require "tmpdir"
+require "byebug"
```
eileencodes: You left byebug in here 😄

lib/licensed/source/bundler.rb
jonabc: @dbussink It's not necessary and won't leak because `::Bundler.with_original_env` captures and restores the full `ENV` object itself.  👍 to restoring the `ENV` var in the `ensure` block to be more explicit and avoid confusion

dbussink: Ah, thanks! Yeah, explicitly doing it is nice I think.

PR #32, opened 2018-05-16
Add pip support for licensed

sridharavinash:
This is an initial take on adding pip package support for `licensed`.

The idea as a start is to `shell` out to `pip` and gather all the information on installed packages. This means we'd have to run this inside a `virutalenv` setting. We also expect a `requirements.txt` that contains all the packages that the current project depends on and could use that to generate the package list and look for the respective license files in the location provided. 

Alternatively `pypi` does have an API that we could use to gather license and other metadata for a package in `requirements.txt` but does not have the `location` of where the files are installed locally (which we'll need to gather the license files itself).


lib/licensed/source/pip.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+
+      def enabled?
+        @config.enabled?(type) && File.exist?(@config.pwd.join("requirements.txt"))
+      end
+
+      def dependencies
+        packages = parse_requirements_txt
+
+        @dependencies = packages.map do |package_name|
```
jonabc: please use `||=` and call `parse_requirements_text` inline so the dependencies will be cached

`@dependencies ||= parse_requirements_txt.map do |package_name|`

lib/licensed/source/pip.rb
```diff
@@ -0,0 +1,64 @@
...
+
+module Licensed
+  module Source
+    class Pip
+      def initialize(config)
+        @config = config
+      end
+
+      def type
+        "python"
```
jonabc: "pip"?

lib/licensed/source/pip.rb
```diff
@@ -0,0 +1,64 @@
...
+              "version"     => package["Version"]
+            })
+        end
+      end
+
+      # Build the list of packages from a 'requirements.txt'
+      # Assumes that the requirements.txt follow the format pkg=1.0.0 or pkg==1.0.0
+      def parse_requirements_txt
+        packages = []
+        File.open(@config.pwd.join("requirements.txt")).each do |line|
```
jonabc: nit: could use `map` here instead of `each` to get rid of the declaration and usage of `packages`

lib/licensed/source/pip.rb
```diff
@@ -0,0 +1,64 @@
...
+          p_split = line.split("=")
+          packages.push(p_split[0])
+        end
+        packages
+      end
+
+      def package_info(package_name)
+        info = {}
+        p_info = pip_command(package_name).split("\n")
+        p_info.each do |pkg|
```
jonabc: nit: could use `each_with_object` here to be a little more concise and not need to externally declare or return `info`

lib/licensed/source/pip.rb
```diff
@@ -0,0 +1,64 @@
...
+          k, v = pkg.split(":", 2)
+          info[k&.strip] = v&.strip
```
jonabc: we probably don't want any nil or empty strings as keys.  can you add a `next if k.nil? || k.empty?`

jonabc: are the calls to `&.strip` due to the lines containing "\n" line endings?  if so it would make more sense to do `pip_command(package_name).split("\n").map(:&strip)` to remove them before evaluating the string content

lib/licensed/source/pip.rb
```diff
@@ -0,0 +1,64 @@
...
+        File.open(@config.pwd.join("requirements.txt")).each do |line|
+          p_split = line.split("=")
+          packages.push(p_split[0])
+        end
+        packages
+      end
+
+      def package_info(package_name)
+        info = {}
+        p_info = pip_command(package_name).split("\n")
```
jonabc: super nit: `.lines` is a ruby-er `.split("\n")`

lib/licensed/source/pip.rb
sridharavinash: 👍

lib/licensed/source/pip.rb
sridharavinash: 👍

lib/licensed/source/pip.rb
sridharavinash: 👍

lib/licensed/source/pip.rb
sridharavinash: `strip` because some of the packages have whitespace when parsing the output of the shelled out command, which messes up the `location` building.

lib/licensed/source/pip.rb
sridharavinash: 👍

sridharavinash:
Testing this on a simple python flask app:

```
root@9e7587f53f1b:/licensed/workspace# licensed list
Displaying dependencies for my_app
  pip dependencies:
    Found Flask (0.12.1)
    Found Flask-Migrate (2.0.3)
    Found Flask-SQLAlchemy (2.2)
    Found Flask-Script (2.0.5)
    Found Jinja2 (2.9.6)
    Found Mako (1.0.6)
    Found MarkupSafe (1.0)
    Found SQLAlchemy (1.1.9)
    Found Werkzeug (0.12.1)
    Found alembic (0.9.1)
    Found click (6.7)
    Found google-api-python-client (1.6.2)
    Found gunicorn (19.7.1)
    Found httplib2 (0.10.3)
    Found itsdangerous (0.24)
    Found oauth2client (4.0.0)
    Found psycopg2 (2.7.1)
    Found pyasn1 (0.2.3)
    Found pyasn1-modules (0.0.8)
    Found python-dateutil (2.6.0)
    Found python-editor (1.0.3)
    Found rsa (3.4.2)
    Found six (1.10.0)
    Found uritemplate (3.0.0)
  * pip dependencies: 24
```
```
licensed status
Checking licenses for my_app: 24 dependencies
FFFFFFFFFFFFFFF.FF.FFFFF
```

sridharavinash:
@jonabc yeah there's where I'm planning on going next .

lib/licensed/source/pip.rb
```diff
@@ -0,0 +1,62 @@
...
+        p_info = pip_command(package_name).lines
+        p_info.each_with_object(Hash.new(0)) { |pkg, a|
+          k, v = pkg.split(":", 2)
+          next if k.nil? || k.empty?
+          a[k.strip] = v&.strip
+        }
+      end
+
+      def pip_command(*args)
+        venv_dir = @config.dig("python", "virtual_env_dir")
```
jonabc: You'll want to root this using `File.expand_path(venv_dir, Licensed::Git.repository_root)`.

script/source-setup/pip
```diff
@@ -0,0 +1,21 @@
...
+  exit 127
+fi
+
+if [ -z "$(which virtualenv)" ]; then
+    echo "A local virtualenv installation is required for python development." >&2
+    exit 127
+fi
+
+
+# setup test fixtures
```
jonabc: can you add support for a "-f" flag that cleans up any existing fixtures before doing the setup steps?  e.g. something like

https://github.com/github/licensed/blob/d22011c9dd162593919752410a642696c9ade454/script/source-setup/go#L14-L16

docs/sources/pip.md
```diff
@@ -0,0 +1,11 @@
+# Pip
+
+The pip source uses `pip` CLI commands to enumerate dependencies and properties. It is expected that `pip` is available in the PATH before running `licensed`
```
jonabc: Is `pip` being available on the PATH needed?  It looks like the source uses `pip` from the virtual env.

jonabc: Can you list the `requirements.txt` file as a requirement for the source being enabled?

docs/sources/pip.md
```diff
@@ -0,0 +1,11 @@
...
+
+#### virtual_env_dir
+A `virtualenv` is assumed to be setup before running `licensed`. The `pip` command will be sourced from this directory.
```
jonabc: > A `virtualenv` is assumed to be setup before running `licensed`.

Can you move ☝️ into the top section as another requirement/expectation?

jonabc: Also can you make note that the `virtual_env_dir` setting is used to find the virtualenv,  that it's a relative path to the repository root, and that the setting is required for the source to work properly?

lib/licensed/source/pip.rb
sridharavinash: 👍

script/source-setup/pip
sridharavinash: 👍

docs/sources/pip.md
sridharavinash: >Is pip being available on the PATH needed? It looks like the source uses pip from the virtual env.

jonabc: > I was thinking about a case where this could be run without a virtual env set

🤔 I guess in that case, you could instead have the configuration setting point to the pip binary location than the virtual env folder.  Then, if the configuration setting is blank licensed would look for `pip` rather than `configured/path/to/pip`

lib/licensed/command/cache.rb
```diff
@@ -11,7 +11,7 @@ def initialize(config)
       def run(force: false)
         summary = @config.apps.flat_map do |app|
           app_name = app["name"]
-          @config.ui.info "Caching licenes for #{app_name}:"
+          @config.ui.info "Caching licenses for #{app_name}:"
```
sridharavinash: looks like @lildude caught the same thing I did 😆 .

docs/sources/pip.md
sridharavinash: @jonabc Thinking about this a little more last night, I think I'd like to leave it as is and have the `venv` directory be the dependency as opposed to the pip binary. The reasons being
* `venv` workflow is pretty standard
* Having everything setup via the virtualenv eliminates the need for the script to handle `pip` args etc
* I don't think `pip` needs to be in the `PATH` since having the `venv` dir specified will derive where the pip binary is located and that should be sufficient.

docs/sources/pip.md
```diff
@@ -0,0 +1,22 @@
...
+pip freeze > requirements.txt
+```
+
+A `virtualenv` is assumed to be setup before running `licensed`. You can setup a `virtualenv` by running the command:
+```
+virtualenv <your_venv_dir>
+```
+_note_: `<your_venv_dir>` path should be relative to the repository root.
+
+#### virtual_env_dir
```
jonabc: nit: not blocking, but I still think this could be more clear that the `virtual_env_dir` setting is required and can be either a relative or absolute path, with the relative path being evaluated from the repository root.

.travis.yml
```diff
@@ -58,5 +58,14 @@ matrix:
       script: ./script/test manifest
       env: NAME="manifest"
 
+    # python tests
+    - language: python
+      python:
+        - "2.7"
+        - "3.6"
```
jonabc: looking at the CI results, it looks like the tests are only run for python v2.7.  I think that due to the way licensed uses an inclusion matrix, and the solution in this file for testing multiple versions of Go, you need to specify the test suite independently for each python version.

sridharavinash: 👍

Issue #28, opened 2018-04-06
Better detection of enabled sources in configuration file


jonabc:
The configuration file can specify whether sources are enabled or disabled with settings like:
```
sources:
  rubygem: true
  go: false
  ...
```

Currently, if a source is not specified it will default to enabled.  A source can individually signal that it's not available via it's own `#enabled?` method, however a different paradigm on the configuration value may be a good idea.

Proposal:

I think the following set of rules would give users better flexibility in specifying which sources they want to run.
1. All sources are enabled via configuration when the `sources` configuration property is not set
2. All sources are enabled via configuration when the `sources` configuration property is empty
3. If any source is enabled via a `true` value, all sources without a `true` value are disabled
```yaml
sources:
  rubygem: true
  npm: false
  # go: false implicitly
```
4. If sources only contains `false` property values, all other sources are enabled
```yaml
sources:
  rubygem: false
  npm: false
  # go: true implicitly
```

The change here is in rule (3), where explicitly setting a `true` value implicitly disables all other sources.  This gives users the option of specifying only the sources they want to run without having to keep an expansive list of all sources that should not run.  It also keeps a good level of flexibility for users to not have to specify every source that should be enabled in every configuration file.

PR #33, opened 2018-05-17
Smarter enabled sources configuration

jonabc:
Fixes https://github.com/github/licensed/issues/28

This follows the plan outlined in https://github.com/github/licensed/issues/28 to determine whether sources are enabled or not from the configuration file:
1. If no sources are configured, all sources are enabled
2. If no sources are explicitly enabled in the configuration, all sources not explicitly configured are enabled
3. If any sources are explicitly enabled in the configuration, all sources not explicitly configured are disabled

The goal is to make it easier and more intuitive to restrict which sources are enabled via configuration.  For most people leaving the setting blank or unconfigured and relying on auto-detection should be 👍 

### TODO:
- [x] Update documentation


jonabc:
Some lesser changes in this PR:
- moved the cabal test fixtures from `test/fixtures/haskell` to `test/fixtures/cabal` to match the source name
- changed the cabal test fixture dependency from `text` to `zlib` due to `zlib` having a broader range of `base` version compatibility
- `DependencySource#type` instance methods are now class methods `DependencySource.type`
- The full list of available source types is stored in `Licensed::AppConfiguration::SourceTypes`

jonabc:
This is fairly non-functional, house cleaning type work.  Merging

jonabc:
@sridharavinash FYI this will impact your current PR for the pip source.

The `#type` dependency source instance method was made into a class method, with usages updated accordingly.  I imagine the changes made to the other sources would be directly applicable to the pip source.  Let me know if you have any questions

PR #34, opened 2018-05-17
Add -f (force) option to setup scripts

jonabc:
Adding force options to setup scripts has been helpful to make sure that test fixtures are 👍 after switching between branches.

This PR also changes the `rake setup` task to run setup scripts using a clean bundler environment.  This fixes a problem where `script/source-setup/bundler` wasn't working properly because it was picking up the parent processes bundler configuration

### TODO
- [x] docs updates


jonabc:
This is a house-cleaning PR and should be a non-functional change, merging

PR #35, opened 2018-05-21
Don't include vendored parts of the golang standard library in packages

WillAbides:
Licensed recognizes `vendor/golang_org/x/text/unicode/norm` as part of the golang standard library, but not when it's vendored as `<pkg>/vendor/golang.org/x/text/unicode/norm`.  This PR will endeavor to correct that.

See https://github.com/golang/go/issues/17326 and https://github.com/golang/go/issues/23802 for why `go list` treats it so.


lib/licensed/source/go.rb
```diff
@@ -61,6 +61,11 @@ def packages
           .uniq
           .select { |d| !go_std_packages.include?(d) }
           .select { |d| !d.start_with?(root_package["ImportPath"]) || vendored_path?(d) }
+          .select do |d|
+            go_std_packages.none? do |std_pkg|
```
jonabc: would you mind adding a comment explaining this / why it's necessary? 🙇

WillAbides:
This is ready for review.  I changed up the test this adds to use a library that was included with golang 1.7 because that's what is being tested on travis.  I also added Gopkg.toml and Gopkg.lock files to the golang test fixtures so that the goland vendor folder can be managed with dep.

For troubleshooting, I added and removed golang 1.10 to .travis.yml, then I created a https://github.com/github/licensed/pull/36 for it to minimize the changes made here.

jonabc:
@WillAbides It looks like all the new vendor files use `golang.org` and not `golang_org`.  Does this verify the new logic?

PR #37, opened 2018-05-22
Add source for go dep

jonabc:
This adds a source for [Go dep](https://github.com/golang/dep) as a dependency package manager.  The big win here is that it removes all shell calls to `go list` making this much faster than the base `Go` dependency source.

Some notable points:
1. This assumes that you are vendoring ALL dependencies, and will self-disable if the source detects any `ignore`d packages  in `Gopkg.toml`
   - if _really_ desired, but not recommended, you can force the source to run via configuration
2. This source does not have a hard dependency on `go` in the local environment
   - if `go` is available, the go std libraries will be filtered out
   - if `go` is not available then `go list std` can't be run and no filtering will happen.  those libraries can still be ignored via configuration.
3. Due to not running `go list`, combined with the lack of summary information available in the `Gopkg.lock` file, there is no summary information available for `dep` dependencies
4. Because the package manager tracks a remote revision SHA of the dependency, the version given in the license file is publicly relevant 🎉 

## TODO
- [x] docs
- [x] CI integration


jonabc:
> Due to not running `go list`, combined with the lack of summary information available in the Gopkg.lock file, there is no summary information available for dep dependencies

/cc @mlinksva the package summary is a nice-to-have but not required from a legal compliance perspective, correct?

mlinksva:
@jonabc 👍 assuming by summary you mean a package's brief self-description analogous to https://guides.rubygems.org/specification-reference/#summary

jonabc:
@mlinksva yep, thats the one.  Thanks!

PR #38, opened 2018-05-24
Raise more informative message after shelling out fails

mislav:
Before:

```sh
$ licensed status -c path/to/config.yml
MYPROJECT/vendor/gems/2.4.2/ruby/2.4.0/gems/json-2.1.0/lib/json/common.rb:156:in `parse': 765: unexpected token at '' (JSON::ParserError)
        from MYPROJECT/vendor/gems/2.4.2/ruby/2.4.0/gems/json-2.1.0/lib/json/common.rb:156:in `parse'
        from MYPROJECT/vendor/gems/2.4.2/ruby/2.4.0/gems/licensed-1.0.1/lib/licensed/source/npm.rb:28:in `dependencies'
```

The trouble is that `npm list` resulted in an error, but we can't see the command executed nor its error output. Instead there is a misleading JSON parse error because it tried to parse an empty string.

After:

```sh
$ licensed status -c path/to/config.yml
/Users/mislav/github/licensed/lib/licensed/shell.rb:93:in `raise_on_failure!': command exited with status 1 (Licensed::Shell::Error)
  npm list --parseable --production --long
    npm ERR! missing: d3-format@1.2.0, required by @github/d3@4.10.2
    npm ERR! missing: d3-geo@1.6.4, required by @github/d3@4.10.2
    npm ERR! missing: d3-zoom@1.5.0, required by @github/d3@4.10.2
        from /Users/mislav/github/licensed/lib/licensed/source/npm.rb:28:in `dependencies'
        from /Users/mislav/github/licensed/lib/licensed/command/status.rb:18:in `each'
```

This extends the `Licensed::Shell` implementation so it raises a helpful message when shelling out fails.

As a bonus, it also switches to a streaming implementation for commands that process output per line.

### DIFF
--- a/lib/licensed/shell.rb
+++ b/lib/licensed/shell.rb
@@ -3,12 +3,19 @@ require "open3"
 
 module Licensed
   module Shell
-    # Executes a command, returning it's STDOUT on success.  Returns an empty
-    # string on failure
-    def self.execute(cmd, *args)
-      output, _, status = Open3.capture3(cmd, *args)
-      return "" unless status.success?
-      output.strip
+    # Executes a command, returning its standard output on success. On failure,
+    # it raises an exception that contains the error output, unless
+    # `allow_failure` is true, in which case it returns an empty string.
+    def self.execute(cmd, *args, allow_failure: false)
+      stdout, stderr, status = Open3.capture3(cmd, *args)
+
+      if status.success?
+        stdout.strip
+      elsif allow_failure
+        ""
+      else
+        raise Error.new([cmd, *args], status.exitstatus, stderr)
+      end
     end
 
     # Executes a command and returns a boolean value indicating if the command
@@ -23,6 +30,32 @@ module Licensed
     def self.tool_available?(tool)
       output, err, status = Open3.capture3("which", tool)
       status.success? && !output.strip.empty? && err.strip.empty?
+    end
+
+    class Error < RuntimeError
+      def initialize(cmd, status, stderr)
+        super()
+        @cmd = cmd
+        @exitstatus = status
+        @output = stderr
+      end
+
+      def message
+        output = @output.to_s.strip
+        extra = output.empty?? "" : "\n#{output.gsub(/^/, "    ")}"
+        "command exited with status #{@exitstatus}\n  #{escape_cmd}#{extra}"
+      end
+
+      def escape_cmd
+        @cmd.map do |arg|
+          if arg =~ /[\s'"]/
+            escaped = arg.gsub(/([\\"])/, '\\\\\1')
+            %("#{escaped}")
+          else
+            arg
+          end
+        end.join(" ")
+      end
     end
   end
 end
--- a/lib/licensed/source/go.rb
+++ b/lib/licensed/source/go.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 require "json"
 require "English"
+require "pathname"
 
 module Licensed
   module Source
@@ -113,7 +114,7 @@ module Licensed
       # package - Go package import path
       def package_info_command(package)
         package ||= ""
-        Licensed::Shell.execute("go", "list", "-json", package)
+        Licensed::Shell.execute("go", "list", "-json", package, allow_failure: true)
       end
 
       # Returns the info for the package under test
@@ -140,7 +141,12 @@ module Licensed
         @gopath = if path.nil? || path.empty?
                     ENV["GOPATH"]
                   else
-                    File.expand_path(path, Licensed::Git.repository_root)
+                    root = begin
+                             Licensed::Git.repository_root
+                           rescue Licensed::Shell::Error
+                             Pathname.pwd
+                           end
+                    File.expand_path(path, root)
                   end
       end
 
--- a/lib/licensed/git.rb
+++ b/lib/licensed/git.rb
@@ -23,7 +23,7 @@ module Licensed
         file = File.directory?(descriptor) ? "." : File.basename(descriptor)
 
         Dir.chdir dir do
-          Licensed::Shell.execute("git", "rev-list", "-1", "HEAD", "--", file)
+          Licensed::Shell.execute("git", "rev-list", "-1", "HEAD", "--", file, allow_failure: true)
         end
       end
 
--- a/lib/licensed/source/cabal.rb
+++ b/lib/licensed/source/cabal.rb
@@ -125,7 +125,7 @@ module Licensed
       # Runs a `ghc-pkg field` command for a given set of fields and arguments
       # Automatically includes ghc package DB locations in the command
       def ghc_pkg_field_command(id, fields, *args)
-        Licensed::Shell.execute("ghc-pkg", "field", id, fields.join(","), *args, *package_db_args)
+        Licensed::Shell.execute("ghc-pkg", "field", id, fields.join(","), *args, *package_db_args, allow_failure: true)
       end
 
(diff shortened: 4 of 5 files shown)


lib/licensed/shell.rb
```diff
@@ -2,27 +2,96 @@
...
+
+    def exit_code
+      @status.exitstatus
+    end
+
+    # Yields each line as it's being read.
+    def each_line
+      return to_enum(__method__) unless block_given?
+
+      @status = Open3.popen3(*@cmd) do |stdin, stdout, stderr, thread|
```
jonabc: it's a little strange that the command is executed only when calling `#each_line`.  Perhaps it should be executed whenever a "terminal" method is called, e.g. `success?`, `exit_code` etc

lib/licensed/shell.rb
```diff
@@ -2,27 +2,96 @@
...
+        thread.value
+      end
+    end
+
+    # Collects all output as a string.
+    def to_s
+      each_line.to_a.join("")
+    end
+
+    def raise_on_failure!
```
jonabc: Could this be a property/attribute on the class that causes a failure to be raised when the command is executed?

e.g. `def initialize(*cmd, raise_on_failure: false)` with pass throughs to the class methods: `def self.execute(cmd, *args, raise_on_failure: false)`

lib/licensed/shell.rb
```diff
@@ -2,27 +2,96 @@
...
+          else
+            arg
+          end
+        end.join(" ")
+      end
+    end
+
+    attr_reader :error_output
+
+    def initialize(*cmd)
```
jonabc: can we make this match the `execute` signature -> `initialize(cmd, *args)` so that it's clear that there needs to be a command and only the args are optional?

lib/licensed/shell.rb
mislav: @jonabc I've defined `each_line` as a “kicker” method here since that's also the method that accepts a block to be yielded in line-by-line streaming mode. If another method kicked off the process, how would you envision `each_line` accepting a block?

Thanks for your comments! I've addressed other concerns in the meantime.

jonabc:
@mislav I'm not sure we need the complexity of a newly added class to describe commands.  My earlier misgiving was that `Licensed::Shell.new(...).each_line` was not immediately clear that it would run the command, or that `Licensed::Shell.new(...).success?` would not.

The ✨ benefits if I understand correctly are:
1. specify if errors should be raised instead of ignoring output when a shell command fails
2. output can given as an enumerable

These benefits only apply to cases where we're concerned with the shell output.  `tool_available?` and `success?` don't see much benefit which means we're talking exclusively about `execute`.

You can get away with all of ☝️without turning `Licensed::Shell` into a full class by updating `Licensed::Shell.execute` to more closely resemble `Licensed::Shell#each_line`.  It would need to handle the `raise_on_failure` flag and behave properly based on whether a block was given or not.
1. block passed in -> behave like `Cmd#each_line`, 
2. block not passed in -> capture all output and return at end

Does that make sense or am I missing a case where `Licensed::Shell` as a class provides additional value?

mislav:
> Does that make sense or am I missing a case where `Licensed::Shell` as a class provides additional value?

These are all good points. Here's how I envisioned that creating instances of Licensed::Shell class might improve things:

1. Explicitly using `each_line` vs. `to_s` instance methods makes call points being explicit regarding what mode of reading output are they opting in to. In terms of code readability, I see this as slightly clearer than the approach of `Licensed::Shell.execute` behaving very differently based on whether a block was passed or not.

2. The class instance allows us to ask the object for the output and for the `success?` status separately (this applies to cases where `raise_on_error: false`). If `Licensed::Shell.execute` was the main and only API, how would we collect all the output of a shell invocation as a string while also checking for exit status? We would need to return two values.

I'm not so hung up on 1 (I could go either way) but I'd love to hear your suggestions about 2. I'm fully open to the fact that I might have overengineered the PR in its current state. However, I do see value in creating a way to shell out to commands that is versatile and has these features:

* It can collect all output as a string
* It can process streaming output line by line
* It can make the exit status available for querying
* It can optionally raise an exception on non-zero exit status

jonabc:
> Explicitly using each_line vs. to_s instance methods makes call points being explicit regarding what mode of reading output are they opting in to. In terms of code readability, I see this as slightly clearer than the approach of Licensed::Shell.execute behaving very differently based on whether a block was passed or not.

If `Licensed::Shell.execute` had a constant return signature would that be more clear? e.g. it always returns `.to_s`, but if you pass a block in it also behaves like `each_line`.

> I'd love to hear your suggestions about 2

@mislav the `Licensed::Shell` module was created to give a simple way to abstract away `Open3`.  The changes here take a step away from the simpleness.  As an example

https://github.com/github/licensed/blob/fd2facbb89c6550989c0f9b15fd1600afa59f7fc/lib/licensed/shell.rb#L14-L18

Previously, this was an abstraction of `Open3`.  Now it's an abstraction of an abstraction of `Open3` with no clear benefits, which signals that this PR makes things more complicated.  Specifically, the abstraction is needed because `Licensed::Shell.new(...).to_s` (`each_line`)  _must_ be called to get the status.  If we return `.success?` and `.tool_available?` back to their previous implementations to maintain simplicity then the only thing the class readily provides is enumerability of output as it's read.

> If Licensed::Shell.execute was the main and only API, how would we collect all the output of a shell invocation as a string while also checking for exit status?

From what I can tell though, we don't have any need for this as `licensed` exists today.  The nature of `licensed` is that any shell commands we run should work, predicated that the tool is available and the environment is properly set up.  As you pointed out in the OP, an `execute` failure is a true failure and should be handled as such, e.g. `raise_on_failure` defaulting to true.

mislav:
@jonabc Thank you for your comments. I have force-pushed a simplified version of the whole approach. `Shell.execute` raises by default, which can be opted-out of by setting `allow_failure: true`.

lib/licensed/shell.rb
```diff
@@ -3,12 +3,19 @@
...
-    # Executes a command, returning it's STDOUT on success.  Returns an empty
-    # string on failure
-    def self.execute(cmd, *args)
-      output, _, status = Open3.capture3(cmd, *args)
-      return "" unless status.success?
-      output.strip
+    # Executes a command, returning its standard output on success. On failure,
+    # it raises an exception that contains the error output, unless
+    # `allow_failure` is true, in which case it returns an empty string.
+    def self.execute(cmd, *args, allow_failure: false)
```
jonabc: ❤️ to raise errors by default

lib/licensed/source/go.rb
```diff
@@ -140,7 +141,12 @@ def gopath
         @gopath = if path.nil? || path.empty?
                     ENV["GOPATH"]
                   else
-                    File.expand_path(path, Licensed::Git.repository_root)
+                    root = begin
```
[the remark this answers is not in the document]
mislav: I could revert this change, but then the tests start failing because this code path is invoked in a context when the `pwd` is in a temporary directory which is not a git repo. This is when I made the call that I will define a fallback here.

Other places that use `repository_root` really depend on it being a git repo, but this one seems like it wants to be able to swing both ways.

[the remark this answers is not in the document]
jonabc: Ah gotcha 👍.  I realized this is equivalent to legacy behavior, though I'm not sure that whether the legacy behavior itself is buggy 😆.

🚢  it!

jonabc:
Everything looks ✨ , merging

PR #43, opened 2018-06-02
Fix test failures when a tool is not available locally

jonabc:
When `script/test` is run without any arguments, `cache_test.rb` and `list_test.rb` each run their respective command for a single test case over all sources.

As part of the release process, I ran `script/test` and found that I was having test failures due to not having a python development environment available with `pip` and `venv`.  The source-specific test files are gated to check that their respective tools are available before the tests are run, however the cache and list tests are somewhat generic, making it hard to check for specific dependencies.

Luckily, the sources already do this with the `#enabled?` method.

This PR changes the `pip` source to check that the expected pip executable exists as part of `#enabled?`.

The relevant test cases have also been updated to check that sources are enabled before trying to use them, and skip the test case if the source is not enabled.

### DIFF
--- a/lib/licensed/source/pip.rb
+++ b/lib/licensed/source/pip.rb
@@ -14,7 +14,8 @@ module Licensed
       end
 
       def enabled?
-        Licensed::Shell.tool_available?(virtual_env_pip) && File.exist?(@config.pwd.join("requirements.txt"))
+        return unless virtual_env_pip && Licensed::Shell.tool_available?(virtual_env_pip)
+        File.exist?(@config.pwd.join("requirements.txt"))
       end
 
       def dependencies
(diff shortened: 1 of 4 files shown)


jonabc:
/cc @sridharavinash FYI I made some changes to the pip source because there were failures in tests where pip was not available.

I added a check that virtual_env_dir pip executable is available to `Licensed::Source::Pip#enabled?`. 
 When the executable isn't found, `enabled?` will return false and the source won't be used.

There is a functional change here - previously a missing `virtual_env_dir` configuration value would raise an error.  I had to remove the error because that code is now run from `enabled?`, meaning that the code will still be executed for projects that are not intending to use the source.

Let me know if you have any concerns about this.  I'd like to get the test failures fixed before releasing the next version of licensed.

I'm guessing others might have seen similar failures from `cache_test.rb` and `license_test.rb`, but `pip` is the first source where I haven't had all the dependencies installed locally 😆

PR #44, opened 2018-06-04
Remove check for [ignored] property for dep source

jonabc:
Spent some time this weekend working on a Golang project that used `dep`, and I realized that the ignored property in `Gopkg.toml` didn't work  the way I thought it did.  Updating the source to remove that check.

The idea behind `dep` is that all dependencies should be in the file, except potentially the go standard library.  As a result, I'm taking `Gopkg.lock` as the source of truth for the v1 release.  If a project relies on `go` packages in the local environment that are not listed in `Gopkg.lock` then the project must still use the `go` source over the `dep` source.

### DIFF
--- a/docs/sources/dep.md
+++ b/docs/sources/dep.md
@@ -3,11 +3,8 @@ This source is intended to be used when all of a projects dependencies have been
 
 # Go Dep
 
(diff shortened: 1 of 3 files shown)


jonabc:
Merging.  If there are issues with this change it can be evaluated in the next release

Issue #53, opened 2018-06-15
Bundler source doesn't work from exe built with ruby-packer


jonabc:
##### Background
https://github.com/github/licensed/issues/50 talks about options for creating easily distributable ways of running licensed from non-ruby environments.

Looking specifically at the executable options, I believe [ruby-packer](https://github.com/pmq20/ruby-packer) is a better choice due to it's easier build pipeline and packaging structure; ruby-packers single executable vs traveling-rubys script + ruby + app.

The ruby-packer-built executable contains a squashed file system with ruby + app which leads to...

##### The problem
When licensed is run on the squashed container file system in the executable, the containers rubygems code is used.  In the ruby-packer project, rubygems has local edits that explicitly disallow loading gem paths from outside the container - ([1](https://github.com/pmq20/ruby-packer/blob/e3d28dc56a9afe31e10639b9ae9afa95ec96300b/ruby/lib/rubygems/path_support.rb#L28-L33) [2](https://github.com/pmq20/ruby-packer/blob/e3d28dc56a9afe31e10639b9ae9afa95ec96300b/ruby/lib/rubygems/path_support.rb#L41-L48))

This issue presents itself when using `::Bundler::Definition#specs`.  The method turns un-evaluated specifications (LazySpecification) into materialized specifications (Specification).  When Bundler looks through it's known specification sources, it's sources are evaluated using rubygems paths and can only find information from inside of the ruby-packer executables file system.

This doesn't work well with the bundler source, who's entire job is to load gemspecs which always exist outside of the container file system. 💣 💥 

##### The solution
Bundler does a lot and needs to support a large number of scenarios.  licensed does not - it requires that bundler has already been run and the environment is ready for sourcing dependencies.  In this scenario information is available at known paths.  Given that information, licensed can sidestep using `::Bundler::Definition#specs` entirely.

Licensed should be able to get by with the specification information from `::Bundler::Definition#resolve` combined with `::Bundler#specs_path`.  The latter information isn't built using path information from rubygems, which makes it 👍 to configure for use outside of the ruby-packer executable's file system.

Issue #50, opened 2018-06-14
Evaluate easy distribution of licensed


jonabc:
Easy distribution != ruby.  That said, there are some options available to make it easier to use `licensed` in non-ruby environments

#### docker images
 It's not entirely clear what the best usage of this would be, though a base ruby image + licensed already installed might be useful?

#### ruby-packaged "executable"
The two most promising options I've seen in this space are
1. [ruby-packer](https://github.com/pmq20/ruby-packer)
2. [traveling-ruby](https://github.com/phusion/traveling-ruby)

They both amount to packaging up ruby + `<app>` into an executable.  ruby-packer would be my preferred approach as it seems to have broader support for varying ruby versions and native extensions.

On the flip side, running a ruby-packer-built executable doesn't play nicely with much referencing gem information from outside the executable which is a current requirement of the `rubygems` source.  I don't know if traveling-ruby has the same issue, but either way [switching to CLI](https://github.com/github/licensed/issues/49) should take care of that.

PR #54, opened 2018-06-15
Source bundler dependencies without using ::Bundler::Definition#specs

jonabc:
Fixes https://github.com/github/licensed/issues/53

This PR updates the bundler source to support finding dependencies in the Bundler source without using Bundler's internal specification `materialize` logic.  It largely aims for unchanged function, though there are two changes called out below.

#### Why
In aiming to be able to distribute a packaged executable of `ruby+licensed`, there were issues with the bundler source.  The executable uses a squashed file system containing ruby, licensed and bundler.  When licensed evaluates a source using the bundler object model, it's using the bundler gem available inside the squashed file system rather than the bundler gem on the host.

When using the bundler gem from the squashed file system, it uses gem information also from inside the squashed file system because it doesn't know about the host file system.  There is a fundamental mismatch which makes sense given that we're running bundler from a separate file system and expecting it to find information from the host automagically.

#### How
Custom logic is added to the bundler source to enumerate specifications that is overall simpler and more straightforward to read through and debug despite having more custom logic in the source.

licensed requires that `bundle install` was already successfully run for dependencies to be enumerated.  As a result, we know that all (*) specifications can be found at `::Bundle.specs_path`.  Both `Gem::Specification` and `::Bundler::Specification` support `#full_name` so we can find the specification using `::Bundle.specs_path.join("#{spec.full_name}.gemspec")`.

**(*)** - There are two exceptions
1. The bundler gemspec itself.  Even when bundler is an explicit Gemfile dependency, it still does not put it's own gemspec under `::Bundler.specs_path`.  More on this in a bit
2. Gemfile dependencies from local gemspecs.  In this scenario we actually don't need to do much.  The realized specification can be obtained from the un-realized specification through `spec.source.specs.first`, where `source` represents the gemspec itself.

It's worth calling out that obtaining the bundler `Gem::Specification` is special cased.  To support the self-contained executable scenario from https://github.com/github/licensed/issues/50, licensed can't use the currently loaded `::Bundler` gem API because it references the gem from the self-contained executable.  It has no knowledge of the host machines Bundler state or installed location.  Instead licensed runs bundler as a separate process on the host machine to find the bundler gem path which is then used to find the specification.

#### Functional changes
1. Bundler is no longer included as a dependency unless it's explicitly called out in the Gemfile.  If the gem hasn't been specified in the Gemfile it's considered a development dependency and filtered.
   - /cc @mlinksva Is this 👍 considering it's likely that bundler is generally assumed to be on the machine and doesn't need to be listed in the Gemfile even when it's a runtime dependency.

2. Configuration overrides for groups to exclude from dependencies is allowed.  Previous functionality is maintained with the default values `:development` and `:test`.

### DIFF
--- a/lib/licensed/source/bundler.rb
+++ b/lib/licensed/source/bundler.rb
@@ -19,7 +19,7 @@ module Licensed
       end
 
       def enabled?
-        defined?(::Bundler) && Licensed::Shell.tool_available?("bundle") && lockfile_path && lockfile_path.exist?
+        defined?(::Bundler) && lockfile_path && lockfile_path.exist?
       end
 
       def dependencies
@@ -135,6 +135,12 @@ module Licensed
       def bundler_spec
         # cache this so we run CLI commands as few times as possible
         return @bundler_spec if defined?(@bundler_spec)
+
+        # finding the bundler gem is dependent on having `gem` available
+        unless Licensed::Shell.tool_available?("gem")
+          @bundler_spec = nil
+          return
+        end
 
         # Bundler is always used from the default gem install location.
         # we can use `gem specification bundler` with a clean ENV to
--- a/lib/licensed/source/bundler.rb
+++ b/lib/licensed/source/bundler.rb
@@ -129,20 +129,18 @@ module Licensed
         end
       end
 
+      # Returns a gemspec for bundler, found and loaded by running `gem specification bundler`
       # This is a hack to work around bundler not placing it's own spec at
       # `::Bundler.specs_path` when it's an explicit dependency
       def bundler_spec
         # cache this so we run CLI commands as few times as possible
         return @bundler_spec if defined?(@bundler_spec)
 
-        # set GEM_PATH to nil in the execution environment to pick up host
-        # information.  this is a specific hack for running from a
-        # ruby-packer built executable
-        path = Licensed::Shell.execute("bundle", "show", "bundler", env: { "GEM_PATH" => nil })
-        # get the gemspec path for the given bundler gem path
-        path = File.expand_path("../../specifications/#{File.basename(path)}.gemspec", path)
-
-        @bundler_spec = Gem::Specification.load(path)
+        # Bundler is always used from the default gem install location.
+        # we can use `gem specification bundler` with a clean ENV to
+        # get the system bundler gem as YAML
+        yaml = ::Bundler.with_original_env { Licensed::Shell.execute("gem", "specification", "bundler") }
+        @bundler_spec = Gem::Specification.from_yaml(yaml)
       end
 
       # Build the bundler definition
--- a/lib/licensed/source/bundler.rb
+++ b/lib/licensed/source/bundler.rb
@@ -39,8 +39,8 @@ module Licensed
       # Returns an array of Gem::Specifications for all gem dependencies
       def specs
         # get the specifications for all dependencies in a Gemfile
-        root_dependencies = definition.dependencies.select { |d| include?(d) }
-        root_specs = specs_for_dependencies(root_dependencies).compact
+        root_dependencies = definition.dependencies.select { |d| include?(d, nil) }
+        root_specs = specs_for_dependencies(root_dependencies, nil).compact
 
         # recursively find the remaining specifications
         all_specs = recursive_specs(root_specs)
@@ -59,8 +59,8 @@ module Licensed
 
         results.merge new_specs
 
-        dependency_specs = new_specs.flat_map { |s| specs_for_dependencies(s.dependencies) }
-                                    .compact
+        dependency_specs = new_specs.flat_map { |s| specs_for_dependencies(s.dependencies, s.source) }
+
         return results if dependency_specs.empty?
 
         results.merge recursive_specs(dependency_specs, results)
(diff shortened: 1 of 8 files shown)


mlinksva:
@jonabc I think that's fine. The thing that would make it not fine (IIUC) would be if there are licensed users who have bundler as a non-explicit but actual runtime dependency and therefore ought be checking/documenting its license but wouldn't be with this change. Assuming that bundler is installed without specifying it seems like a bad idea that would eventually bite such a user anyway, thus an edge case that we shouldn't worry about. Sound right?

Also, I looked at the changes here and don't understand well enough to see why without this PR, licensed includes bundler as a dependency even if it is not specified. Just out of curiosity, where is this PR filtering out bundler unless it is explicitly included?

lib/licensed/source/bundler.rb
```diff
@@ -35,14 +36,133 @@ def dependencies
...
+
+      # Returns whether a dependency should be included in the final
+      def include?(dependency, source)
+        # ::Bundler::Dependency has an extra `should_include?`
+        return false unless dependency.should_include? if dependency.respond_to?(:should_include?)
+
+        # Don't return gems added from `add_development_dependency` in a gemspec
+        # if the :development group is excluded
+        gemspec_source = source.is_a?(::Bundler::Source::Gemspec)
+        return false if dependency.type == :development && (!gemspec_source || exclude_development_dependencies?)
```
jonabc: This filters out gems that are added to a gemspec via `add_development_dependency`.  `!gemspec_source` covers dependencies that are found recursively from other dependencies (`source` should be `nil`).  `exclude_development_dependencies?` checks that the `:development` group is excluded via bundler or licensed configuration.

If bundler was listed as a development dependency via a local gemspec, or if it was found recursively as a development dependency of a runtime dependency, it will be filtered here.

lib/licensed/source/bundler.rb
```diff
@@ -35,14 +36,133 @@ def dependencies
...
+
+        # Gem::Dependency don't have groups - in our usage these objects always
+        # come as child-dependencies and are never directly from a Gemfile.
+        # We assume that all Gem::Dependencies are ok at this point
+        return true if dependency.groups.nil?
+
+        # check if the dependency is in any groups we're interested in
+        (dependency.groups & groups).any?
```
jonabc: This filters our gems that are not found in any Gemfile groups that will be included in licensed actions.

If bundler is listed as a dependency under a `:development` group, it'll get filtered here when the `:development` group is excluded either by bundler or licensed configurations.  

By default licensed **does** exclude the `:development` group.

jonabc:
> The thing that would make it not fine (IIUC) would be if there are licensed users who have bundler as a non-explicit but actual runtime dependency and therefore ought be checking/documenting its license but wouldn't be with this change.

Yea, that situation is what I'm concerned about as well. 

> Assuming that bundler is installed without specifying it seems like a bad idea that would eventually bite such a user anyway, thus an edge case that we shouldn't worry about.

👍 Agreed that assuming anything is installed without specifying it is a bad idea.  My concern is that I don't know if it's edge case or if it's common.  Bundler is the de facto package manager for ruby and is installed almost everywhere - it seems pretty easy to assume that it's going to be available at runtime.

> Also, I looked at the changes here and don't understand well enough to see why without this PR, licensed includes bundler as a dependency even if it is not specified.

I've looked through the bundler source a few times and I'm still not sure I understand it.  I think bundler might always add itself with `type == :development` if not otherwise specified in a Gemfile or gemspec.

> Just out of curiosity, where is this PR filtering out bundler unless it is explicitly included?

I added some review comments trying to explain a little better how dependencies get filtered out, with specific examples of how bundler might be filtered when it's a development dependency.

jonabc:
As [CI highlighted](https://travis-ci.org/github/licensed/builds/392939089?utm_source=github_status&utm_medium=notification), bundler doesn't always know where bundler is located.

I switched to using `gem specification` and `Gem::Specification.from_yaml` to create the bundler specification.  This should be safe and reliable - bundler is always used from the current global gem list that is also used for `gem specification`.

It looks like CI is green with that change

jonabc:
And finally, to highlight the reason behind the PR.  With these changes I'm able to build executables for licensed that can be used for all dependency sources, in any environment, whether ruby is present or not 🎉 🎉 

```
➜  ./licensed list -c test/fixtures/config/.licensed.yml
Displaying dependencies for licensed-yml
  rubygem dependencies:
    Found addressable (2.5.2)
    Found bundler (1.16.1)
    Found dotenv (2.4.0)
    Found faraday (0.15.2)
    Found licensee (9.9.1)
    Found multipart-post (2.0.0)
    Found octokit (4.8.0)
    Found pathname-common_prefix (0.0.1)
    Found public_suffix (3.0.2)
    Found rugged (0.27.1)
    Found sawyer (0.8.1)
    Found thor (0.20.0)
    Found tomlrb (1.2.7)
  * rubygem dependencies: 13
```

tenderlove:
> Assuming that bundler is installed without specifying it seems like a bad idea that would eventually bite such a user anyway, thus an edge case that we shouldn't worry about. Sound right?

I agree with this assessment.  In order to `gem install` such a gem, there would have to be *something* in the dependency tree that depends on it (otherwise you just wouldn't be able to install that gem).  Today, the only things that can be implicitly depended upon and *not* installed are things that ship with Ruby's standard library.

jonabc:
@benbalter can you take a look at this?  I'd ❤️ to be able to attach executable packages along with each `licensed` release.

jonabc:
Merging.  Can address any issues if/as they come up.

PR #56, opened 2018-06-21
Workflow and scripting to build and package distributable exes

jonabc:
Fixes https://github.com/github/licensed/issues/50.  Docker images are still interesting to consider however having an executable publicly available to download makes it fairly easy to use licensed from any existing docker image.

This PR introduces scripting for building licensed executables with [`ruby-packer`/`rubyc`](https://github.com/pmq20/ruby-packer).  The default `package` rake task is overridden to run the package scripting.

Building executables for a platform needs to happen on that platform, and this PR uses docker to try and make building packages as cross-platform as possible.  It currently includes packaging scripts for mac and linux.  

Building for windows is not supported.  I don't have a windows environment available to build windows exes, though if this approach is 👍 I'm going to open an issue to investigate if any of the [microsoft docker images](https://hub.docker.com/u/microsoft/) would suffice.  Licensed needs a few tweaks to work on windows anyways so this isn't yet the highest priority.

Building a mac exe needs to happen from MacOS due to (AFAIK) there not being any suitable mac docker images.

Building a linux exe can happen on any environment due to the abundance of linux images.  If a user is already on linux, building the linux exe will happen on the host system.

Note: Executables for versions 1.1.0 and prior will not work for sourcing bundler dependencies - please see https://github.com/github/licensed/pull/54 for a description and fix.

### DIFF
--- a/docker/Dockerfile.build-linux
+++ b/docker/Dockerfile.build-linux
@@ -1,0 +1,10 @@
+FROM ruby:2.4-slim-stretch
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends cmake make gcc pkg-config squashfs-tools git curl bison \
+    && rm -rf /var/lib/apt/lists/* \
+    && curl -L http://enclose.io/rubyc/rubyc-linux-x64.gz | gunzip > /usr/local/bin/rubyc \
+    && chmod +x /usr/local/bin/rubyc
+
+ENV CPPFLAGS="-P"
+ENV RUBYC="/usr/local/bin/rubyc"
--- a/docker/Dockerfile.build-linux
+++ b/docker/Dockerfile.build-linux
@@ -2,8 +2,9 @@ FROM ruby:2.4-slim-stretch
 
 RUN apt-get update \
     && apt-get install -y --no-install-recommends cmake make gcc pkg-config squashfs-tools git curl bison \
-    && rm -rf /var/lib/apt/lists/* \
-    && curl -L http://enclose.io/rubyc/rubyc-linux-x64.gz | gunzip > /usr/local/bin/rubyc \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN curl -L http://enclose.io/rubyc/rubyc-linux-x64.gz | gunzip > /usr/local/bin/rubyc \
     && chmod +x /usr/local/bin/rubyc
 
 ENV CPPFLAGS="-P"
--- a/docs/packaging.md
+++ b/docs/packaging.md
@@ -1,0 +1,40 @@
+# Packaging licensed for distribution
+
+Licensed is built into executables and packaged for distribution using [ruby-packer][ruby-packer].
+
+Executable packages are currently supported for:
+1. Linux
+2. MacOS / Darwin
+
+The packaged executables contain a self-expanding file system containing ruby, licensed and all of it's runtime dependencies.  Licensed is run inside the contained file system, allowing usage in scenarios where ruby is not available on the host system.
+
+### Building packages
+
+Packages are built as `licensed-$VERSION-$PLATFORM-x64.tar.gz` tarballs that contain a single `./licensed` executable.  After building a package through the available scripting, it will be available in the `pkg` directory.
+
+#### Building all packages
+```bash
+# build all packages
+$ script/package
+```
+or
+```bash
+# build all packages
+$ bundle exec rake package
+```
+
+#### Building packages for a single platform
+```bash
+# build package for linux
+$ script/package linux
+```
+or
+```bash
+# build package for linux
+$ bundle exec rake package[linux]
+
+# if using the zsh shell then you'll need to escape the brackets
+$ bundle exec rake package\[linux\]
+```
+
+[ruby-packer]: https://github.com/pmq20/ruby-packer
--- a/docs/packaging.md
+++ b/docs/packaging.md
@@ -36,10 +36,7 @@ $ script/package linux
 or
 ```bash
 # build package for linux
-$ bundle exec rake package[linux]
-
-# if using the zsh shell then you'll need to escape the brackets
-$ bundle exec rake package\[linux\]
+$ bundle exec rake package:linux
 ```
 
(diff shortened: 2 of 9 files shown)


docs/packaging.md
```diff
@@ -0,0 +1,56 @@
...
+`$VERSION` in the package name will be set to the current branch name if
+available, otherwise the current SHA.  To use a specific licensed version,
+set a `VERSION` environment variable when calling the packaging scripts.  `VERSION` can be set to any value that works with `git checkout`.
+
+#### Building all packages
+```bash
+# build all packages
+$ script/package
+```
+or
```
sridharavinash: Is there any advantage on using `bundle` over `script`? I think we can stick with just `script/package` if there's no difference just for the sake of consistency/simplification?

sridharavinash:
>Building for windows is not supported. I don't have a windows environment available to build windows exes, though if this approach is 👍 I'm going to open an issue to investigate if any of the microsoft docker images would suffice. Licensed needs a few tweaks to work on windows anyways so this isn't yet the highest priority.

I think with [Windows Subsystem or Linux](https://docs.microsoft.com/en-us/windows/wsl/install-win10) you can run the `linux` exe without having to change anything.

docs/packaging.md
jonabc: @sridharavinash it's a matter of conventions 😄

[GitHub scripts to rule them all](https://github.com/github/scripts-to-rule-them-all) vs general ruby practices of using rake tasks.  It's not much work to allow users equivalent choices 🤷‍♂️

jonabc:
> I think with Windows Subsystem or Linux you can run the linux exe without having to change anything.

@sridharavinash 👍 as a workaround.

It would still be valuable at some point to have a native windows exe so that additional setup steps aren't needed to have a viable local environment for running licensed.

docs/packaging.md
sridharavinash: 👍  fair enough!

docker/Dockerfile.build-linux
```diff
@@ -0,0 +1,10 @@
+FROM ruby:2.4-slim-stretch
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends cmake make gcc pkg-config squashfs-tools git curl bison \
+    && rm -rf /var/lib/apt/lists/* \
+    && curl -L http://enclose.io/rubyc/rubyc-linux-x64.gz | gunzip > /usr/local/bin/rubyc \
```
tbehling: Just a thought, downloading a binary from a non-TLS URL could make some users uncomfortable.  How widely used is this Dockerfile expected to be used -- just for internal development or will ordinary licensed users run this too?

It looks like `enclose.io` sadly doesn't have a valid TLS cert.  Is there an HTTPS URL this could be downloaded from?

Not a big diff but I'd put the rubyc curl in a separate RUN line so it's cached separately from the apt-get (in fact maybe even before it so the curl will still be cached when the apt-get changes).

docker/Dockerfile.build-linux
```diff
@@ -0,0 +1,10 @@
+FROM ruby:2.4-slim-stretch
```
tbehling: FYI, I see Ruby 2.5.x is stable now if you were interested in moving up.

docs/packaging.md
```diff
@@ -0,0 +1,56 @@
...
+# build package for linux
+$ script/package linux
+```
+or
+```bash
+# build package for linux
+$ bundle exec rake package[linux]
+
+# if using the zsh shell then you'll need to escape the brackets
+$ bundle exec rake package\[linux\]
```
tbehling: This could probably be consolidated if the first line had quotes: `bundle exec rake 'package[linux]'`

script/packages/linux
```diff
@@ -0,0 +1,50 @@
...
+#/ Packages are of the form licensed-$VERSION-linux-x64.tar.gz and contain a `./licensed` executable
+#/ Built packages are placed in the <root>/pkg directory.
+#/
+#/ If calling from a non-linux OS, docker is used to build a linux binary
+#/
+#/ OPTIONS:
+#/   [VERSION]         (optional, default to current git branch or SHA1) version of licensed to build exe at
+#/
+
+set -e
```
tbehling: FYI if you want to go a little further with Bash safety, you could use `set -euo pipefail`, which I've been putting into new scripts.  See https://coderwall.com/p/fkfaqq/safer-bash-scripts-with-set-euxo-pipefail

docker/Dockerfile.build-linux
jonabc: > How widely used is this Dockerfile expected to be used -- just for internal development or will ordinary licensed users run this too?

Only for internal development / building packages.  General usage will download the built packages from a GitHub release which is HTTPS 👍 

> It looks like enclose.io sadly doesn't have a valid TLS cert. Is there an HTTPS URL this could be downloaded from?

😢 I don't think so.  I'll file an issue at the repo to ask them to set up TLS.

docker/Dockerfile.build-linux
jonabc: 👍 , but as a separate PR.  Upgrading to a new ruby version should also update CI etc..

docs/packaging.md
jonabc: True, except it's not (AFAIK) super common to quote the command run by bundle exec.  I'm going to update this to look like `bundle exec rake package:linux` and sidestep the whole thing.

docker/Dockerfile.build-linux
jonabc: > Not a big diff but I'd put the rubyc curl in a separate RUN line so it's cached separately from the apt-get (in fact maybe even before it so the curl will still be cached when the apt-get changes).

👍 to separating them.  I'd expect rubyc to be updated more frequently than any of the apt-get requirements, so it should probably come after.

jonabc:
Additional testing with the help of @sridharavinash (🙇!!) found an issue with the build scripting where the destination directory (`pkg`) might not exist before trying to place a `tar.gz` there.

PR has been updated for feedback and to fix ☝️

PR #57, opened 2018-06-22
Release 1.2.0

jonabc:
### Added
- Building and packaging distributable exes for licensed releases
- Can now configure which Gemfile groups are excluded from dependency enumeration

### Fixed
- Bundler is no longer always reported as a dependency
- Set the minimum required ruby version for licensed


jonabc:
Tested built gem on local project, all 👍

PR #60, opened 2018-06-24
added path property for npm dependencies

kumarmj:
### This PR brings

- `npm.rb` (*addition*)
   - added `path` property for npm dependecies

solves: #4

### DIFF
--- a/lib/licensed/source/npm.rb
+++ b/lib/licensed/source/npm.rb
@@ -19,16 +19,10 @@ module Licensed
       def dependencies
         return @dependencies if defined?(@dependencies)
 
-        locations = {}
-        package_location_command.lines.each do |line|
-          path, id = line.split(":")[0, 2]
-          locations[id] ||= path
-        end
-
         packages = recursive_dependencies(JSON.parse(package_metadata_command)["dependencies"])
 
         @dependencies = packages.map do |name, package|
-          path = package["path"] || package["realPath"] || locations["#{package["name"]}@#{package["version"]}"]
+          path = package["path"]
           fail "couldn't locate #{name} under node_modules/" unless path
           Dependency.new(path, {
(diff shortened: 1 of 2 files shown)


kumarmj:
Node v4.9.1 NPM 2.15.11 contains both "path" & "realPath"
Node v6.11.1 NPM 3.10.10 contains only "path"
Node v8.11.3 NPM 5.6.0 contains only "path"

kumarmj:
Node 4 NPM 2 -> Number of `"json path"` equal `npm list --parseable` output paths
Node 6 NPM 3 -> Number of `"json path"` equal `npm list --parseable` output paths
Node 8 NPM 5 ->  Number of `"json path"` lot more greater than `npm list --parseable` output paths

I used `create-react-app` using `nvm` for environment changing in Node 8 there were 1387 `--parseable paths` where 2464 `--json paths`

kumarmj:
I did not encountered any path in all three `--json outputs` that is `nil` or `empty` all had valid paths

jonabc:
@kumarmj thanks for the additional testing! 🙇 

Following up on your testing, I also tested those workflows with Node 10 / NPM 6 and found similar results as you did for Node 8 / NPM 5

From what I can tell, the `path` property should be safe to use exclusively. 

1. The `realPath` property hasn't been output since Node v4 which is [EOL](https://github.com/nodejs/Release#end-of-life-releases).
2. From a quick look at the [oldest tagged source I could find](https://github.com/npm/npm/blob/6aa71b4b4a2595649b73b7be0b9d8d798f3cd45c/lib/ls.js#L443-L444) it appears that the path licensed parses from `npm --production --long --parseable` is equivalent to the `path` property 👍 

You can go ahead and remove the checks for `realPath` and `locations`  🎉

Issue #59, opened 2018-06-23
[manifest] licensed doesn't find license content from file comment headers


jonabc:
The manifest dependency source is intended to be used in situations where a package manager is not available.  Usages so far have been for c/c++ projects where dependencies are vendored into the project as source copied from another project.

There is significant friction in using this source when vendored source code doesn't contain an explicit `LICENSE` type file.  This is not uncommon for c/c++ projects where license and copyright information is often copied directly into each file as a comment header.  The current workflow for updating dependencies in those projects looks like

1. run licensed
   1. if licensed finds dependency updates, it updates the version and clears out the license text / metadata for each updated dependency
2. re-add license text / metadata as needed for each dependency and commit
3. re-run licensed to ensure everything is up to date 👍 

Neither licensed nor licensee currently have the ability to understand license/copyright information as a comment header.  licensee explicitly states that it's [not a project goal](https://github.com/benbalter/licensee/blob/master/docs/what-we-look-at.md#what-about-checking-every-single-file-for-a-copyright-header) to support header comments, which means that fixing this should be handled in licensed.

PR #61, opened 2018-06-25
[Manifest] Use licenses from source file comments if license is not otherwise found

jonabc:
Fixes https://github.com/github/licensed/issues/59

This PR updates the manifest dependency enumerator to parse unique license content from source file comments when license content is not otherwise found.

The manifest source uses a specialized `Dependency` class that has a backup implementation for finding license content if `license_key` is "none".  When needed, it will pull all unique comments containing `copyright` text into a separate `LICENSE` file which is then used by `licensee` to determine the license.

Questions
1. How should we handle when there is both a LICENSE type file, and license content in source file comments?  Currently this is not looking at source files if a license is found, even when it's "other".
2. `Licensed::Source::Manifest::Dependency` is only checking the license key when determining whether to pull license content from source files.  Do we need to also consider legal notices?
3. License content is being found based on whether a comment contains the word `copyright`.  Is that appropriate, or is there better content to match on that is more indicative of the comment containing licensing content?

### DIFF
--- a/docs/sources/manifests.md
+++ b/docs/sources/manifests.md
@@ -24,3 +24,35 @@ File paths are relative to the git repository root.  Package names will be used 
 If multiple source files map to a single package and they share a common path under the git repository root, that directory will be used to find license information, if available.
 
 It is the responsibility of the repository owner to maintain the manifest file.
+
+### Finding license content from source file comments
+
+When a file containing license content is not found for a group of source files,
+Licensed will attempt to parse license text from source file comments.
+
+There are some limitations on this functionality:
+
+1. Comments MUST contain a copyright statement
+2. Comments MUST be C-style multiline comments, e.g. `/* comment */`
+3. Comments SHOULD contain identical indentation for each content line.
+
+The following examples are all :+1:.  Licensed will try to preserve formatting,
+however for best results comments should not mix tabs and spaces in leading whitespace.
+```
+/*
+   <copyright statement>
+
+   <license text>
+ */
+
+/* <copyright statement>
+   <license text>
+ */
+
+/*
+ * <copyright statement>
+ * <license text>
+ *
+ * <license text>
+ */
+```
--- a/lib/licensed/dependency.rb
+++ b/lib/licensed/dependency.rb
@@ -9,22 +9,27 @@ module Licensed
     attr_reader :search_root
 
     def initialize(path, metadata = {})
-      @path = path
       @search_root = metadata.delete("search_root")
-
-      # with licensee providing license_file[:dir],
-      # enforcing absolute paths makes life much easier when determining
-      # an absolute file path in notices
-      unless Pathname.new(path).absolute?
-        raise "Dependency path #{path} must be absolute"
-      end
-
       super metadata
+
+      self.path = path
     end
 
     # Returns a Licensee::Projects::FSProject for the dependency path
     def project
       @project ||= Licensee::Projects::FSProject.new(path, search_root: search_root, detect_packages: true, detect_readme: true)
+    end
+
+    # Sets the path to source dependency license information
+    def path=(path)
+      # enforcing absolute paths makes life much easier when determining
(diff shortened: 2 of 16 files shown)


benbalter:
@jonabc nice. Wondering if it makes more sense to implement this upstream within Licensee as an additional source? That way it should "just work" in most contexts? Also curious how you'd handle conflicting license information (e.g., LICENSE says A, but files say B, and C)? Same with distinguishing vendored libraries?

jonabc:
> > Also curious how you'd handle conflicting license information (e.g., LICENSE says A, but files say B, and C)?

@benbalter 😆 Great question!

> How should we handle when there is both a LICENSE type file, and license content in source file comments?
> Currently this is not looking at source files if a license is found, even when it's "other".

jonabc:
I'd be happy to make this part of `licensee`.  I implemented it in `licensed` because it seemed like doing this was explicitly [not a goal of licensee](https://github.com/benbalter/licensee/blob/master/docs/what-we-look-at.md#what-about-checking-every-single-file-for-a-copyright-header)

mlinksva:
@jonabc so I understand -- are 1&2 the same question -- what to do with source notices in the presence of a `LICENSE` file or package metadata? Suspect it would be an improvement to go ahead and fix #59 with this PR; more robust handling source notices in all cases is another issue; it might make sense for some or all of the code to be in licensee.

For trivial 3, probably also if it contains `(c)|©|licen[s|c]e`.

benbalter:
[`Licensee::Matchers::Copyright::REGEX`](https://github.com/benbalter/licensee/blob/master/lib/licensee/matchers/copyright.rb#L8) may be useful since we already have the dependency.

jonabc:
> are 1&2 the same question -- what to do with source notices in the presence of a LICENSE file or package metadata?

@mlinksva same general topic but I'm not sure it's the same question.

(1) is checking if it's 👍 to forego source file comment checking when licensee reports that license content was found
(2) is checking if it's 👍 to ignore non-license legal notices when pulling license(s) from source file comments

mlinksva:
@jonabc ah I thought both were (1). Anyway, (1) it would be good to look for conflicting information in source comments, but I think it would be OK for now to only look in source comments when a license can't otherwise be detected, which would address #59. (2) I don't think we need to look for comments that look like potential legal notices but don't look like license or copyright notices.

Regarding the regex for C-style comments -- I imagine it would be possible to have regexes for different comment delimiters and use depending on language of source file in question? But starting off by only supporting C-style would still solve a problem.

jonabc:
@mlinksva 

> (2) I don't think we need to look for comments that look like potential legal notices but don't look like license or copyright notices.

I think I explained this badly.  I was describing a scenario where changing `Dependency#path` to a temp directory discarded legal notices found from the original location.  

I took another look at it, and the scenario has been [fixed](https://github.com/github/licensed/pull/61/commits/15cba7fd66e018491b4f5a01f507836358c6fd38) and [tested](https://github.com/github/licensed/pull/61/commits/bd72e775440b11efdea627789f1d92ca3bcfd772).

> Regarding the regex for C-style comments -- ... But starting off by only supporting C-style would still solve a problem.

👍 I think I agree with that.  Supporting `/* */` comments should cover a large number of popular languages that don't already have explicit package manager support and we can work towards broader support in the future.

Long term, I'm not sure it makes sense to implement detailed comment parsing logic directly in `licensed` for this one specific purpose.  We can cross that bridge when we need to.

jonabc:
As a note, I've updated the manifest source to use whatever length of indentation is set for the first line of the comment that contains word characters as the expected indentation for the rest of the comment.  This lets licensed get away with not fully stripping the leading whitespace and preserving _some_ of the formatting of the license text.

It's not the smartest solution out there, but it should be good enough for the time being.

jonabc:
Moving ahead and merging this, will continue to iterate on comment matching as needed in the future

Issue #22, opened 2018-04-04
Cache all available license files


jonabc:
First mentioned in https://github.com/github/licensed/pull/21#issuecomment-378682909.

When `licensee` finds multiple license files in a project, it doesn't attempt to determine which is the "valid" license content.  As a result the `Project#license_file` method returns nil if more than one license file is found. 

This makes sense for `licensee` but for `licensed`'s purposes, it may be useful to display all license file text.

Some considerations if all license files are written to the cached file
1. Are there any legal consequences to displaying multiple licenses if they are conflicting or it's unclear which one is valid?
1. There may be a mismatch between license content shown and the `license` metadata value
1. How to distinguish where the license content came from
   - Is this even needed?  We don't label where `Dependency#notices` files come from in the cached file

/cc @mlinksva @benbalter

mlinksva:
1. never hurts to cache more of what the project documents about itself regarding licenses and related notices
2. in the case of multiple licenses found, that'll always be `other`, right?
3. not a requirement, but might be nice to include what file(s) licenses/notices are sourced from in the metadata

Licensee does make all this info available.

jonabc:
> in the case of multiple licenses found, that'll always be other, right?

Not necessarily - if the project specifies a license in its package metadata (e.g. gemspec for a rubygems project), or if it's parseable from a readme then those will take precedence over `other`

mlinksva:
Ah, right. It's still safer to include other license texts found, which could be eg for third party material the project includes and for which license notices should still be reproduced for. Of course the license text for the license identified by the project would be best output first in the cached document.

benbalter:
> When `licensee` finds multiple license files in a project, it doesn't attempt to determine which is the "valid" license content. As a result the `Project#license_file` method returns nil if more than one license file is found.

*Technically* that behavior was implemented to preserve backward compatibility. If you call `#license_files` or `#licenses` (plural), it should give you everything it knows (but it's still up to Licensed or a human how to evaluate conflicting license information).

PR #64, opened 2018-07-09
Dependencies cache all available license files

jonabc:
Fixes https://github.com/github/licensed/issues/22.

This PR updates the `licensee` usage to cache all license texts associated with a dependency.  The implementation changes
- `#license_text` to account for using `project.license_files` over `project.license_file`
- `#matched_project_file` -> `project_license`, which checks for (in order)
   1. A non-other licensee `project.license` 👍  
   2. Multiple non-unique license files to match as `other` (should this match to "multiple"?)
   3. Legacy behavior that ends up calling into `project.matched_files...`
- removes `#project_file` and instead moves `#remote_license_file` usage directly into `#license_key`

This is not a breaking change, however it will cause a mismatch between license text cached from `licensed <= 1.2.0` and license text cached `licensed > 1.2.0` for dependencies with multiple license files.  In that scenario, the cached license metadata field will not be reused per https://github.com/github/licensed/pull/21

/cc @benbalter added you as reviewer for changes to how we're using `licensee`

### DIFF
--- a/lib/licensed/dependency.rb
+++ b/lib/licensed/dependency.rb
@@ -68,14 +68,8 @@ module Licensed
     # Resets all local project and license information
     def reset_license!
       @project = nil
-      @project_license = nil
       self.delete("license")
       self.text = nil
-    end
-
-    # Finds and returns a license from the Licensee::FSProject for this dependency.
-    def project_license
-      @project_license ||= project.license
     end
 
     # Returns a Licensee::LicenseFile with the content of the license in the
@@ -96,11 +90,8 @@ module Licensed
 
     # Returns a string representing the project's license
     def license_key
-      if project_license
-        project_license.key
-      else
-        "none"
-      end
+      return "none" unless project.license
+      project.license.key
     end
   end
 end
--- a/lib/licensed/dependency.rb
+++ b/lib/licensed/dependency.rb
@@ -68,24 +68,24 @@ module Licensed
     # Resets all local project and license information
     def reset_license!
       @project = nil
-      @matched_project_file = nil
+      @project_license = nil
       self.delete("license")
       self.text = nil
     end
 
-    # Returns the Licensee::ProjectFile representing the matched_project_file
-    # or remote_license_file
-    def project_file
-      matched_project_file || remote_license_file
-    end
-
-    # Returns the Licensee::LicenseFile, Licensee::PackageManagerFile, or
-    # Licensee::ReadmeFile with a matched license, in that order or nil
-    # if no license file matched a known license
-    def matched_project_file
-      @matched_project_file ||= project.matched_files
-                                       .select { |f| f.license && !f.license.other? }
-                                       .first
+    # Finds and returns a license from the Licensee::FSProject for this dependency.
+    def project_license
+      @project_license ||= if project.license && !project.license.other?
+        # if Licensee has determined a project to have a specific license, use it!
+        project.license
+      elsif project.license_files.map(&:license).uniq.size > 1
+        # if there are multiple license types found from license files,
+        # return 'other'
+        Licensee::License.find('other')
+      else
+        # check other project files if we haven't yet found a license type
+        project.licenses.reject(&:other?).first
+      end
     end
 
     # Returns a Licensee::LicenseFile with the content of the license in the
@@ -96,21 +96,28 @@ module Licensed
     end
 
     # Regardless of the license detected, try to pull the license content
-    # from the local LICENSE, remote LICENSE, or the README, in that order
+    # from the local LICENSE-type files, remote LICENSE, or the README, in that order
     def license_text
-      content_file = project.license_file || remote_license_file || project.readme_file
-      content_file.content if content_file
+      content_files = Array(project.license_files)
+      content_files << remote_license_file if content_files.empty? && remote_license_file
+      content_files << project.readme_file if content_files.empty? && project.readme_file
+      content_files.map(&:content).join("\n#{LICENSE_SEPARATOR}\n")
     end
 
     # Returns a string representing the project's license
-    # Note, this will be "other" if a license file was found but the license
-    # could not be identified and "none" if no license file was found at all
     def license_key
-      if project_file && project_file.license
-        project_file.license.key
-      elsif project.license_file || remote_license_file
+      if project_license && !project_license.other?
+        # return a non-other license found from the Licensee
+        project_license.key
+      elsif remote_license_file && !remote_license_file.license.other?
+        # return a non-other license found from the remote source
+        remote_license_file.license.key
+      elsif project.license || remote_license_file
+        # if a license was otherwise found but couldn't be identified to a
+        # single license, return "other"
         "other"
       else
+        # no license found
         "none"
       end
     end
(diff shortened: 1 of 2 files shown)


lib/licensed/dependency.rb
```diff
@@ -68,24 +68,24 @@ def local_files
...
-                                       .first
+    # Finds and returns a license from the Licensee::FSProject for this dependency.
+    def project_license
+      @project_license ||= if project.license && !project.license.other?
+        # if Licensee has determined a project to have a specific license, use it!
+        project.license
+      elsif project.license_files.map(&:license).uniq.size > 1
+        # if there are multiple license types found from license files,
+        # return 'other'
+        Licensee::License.find('other')
```
mlinksva: Could the elsif branch be deleted if `&& !project.license.other?` were also deleted?

lib/licensed/dependency.rb
```diff
@@ -96,21 +96,28 @@ def remote_license_file
     end
```
mlinksva: Not part of this change, but just idly wondering why above two lines can't just be

`@remote_license_file ||= Licensed.from_github(self["homepage"])`

jonabc: Good question!  It's because `||=` doesn't stop `Licensed.from_github(self["homepage"])` from being called again when a previous call returned `nil`

lib/licensed/dependency.rb
jonabc: I don't think it can.  The call chain for  callstack for this would look like
- `Project#license`
- `Project#licenses_without_copyright`
- `Project#matched_files`
- `Project#project_files`

`#project_files` pulls from `#license_files`, `#package_manager_file` and `#readme_file`.  For the purposes of this PR I was calling out a special case when there were multiple, conflicting, license files.

Note that this doesn't include a scenario where there are multiple license files but they're all other.  In that scenario there is still only a single license found from license files and we fall into the `else` case to look at the package manager and readme files.

mlinksva: > For the purposes of this PR I was calling out a special case when there were multiple, conflicting, license files.

Doesn't licensee handle that case directly by itself?

https://github.com/benbalter/licensee/blob/d606b06860fd290693280060faee9c61cbf41c4b/lib/licensee/projects/project.rb#L26-L27

jonabc: It's really similar, but not exactly the same.  What you linked is evaluating licenses from license files, package manager files and readme files.  The one in my change is only looking at [license files](https://github.com/benbalter/licensee/blob/d606b06860fd290693280060faee9c61cbf41c4b/lib/licensee/project_files/license_file.rb#L26-L43), where I use the term to mean files matching that regex e.g. LICENSE or COPYING.

Before this change, `licensed` used the first `project.matched_files` file that had a non-other license.  The issue I saw was that when using multiple license files, is it ok to say that one of them is more relevant than another?

For instance if there are both `LICENSE` and `COPYING` files that contain different licenses, is it ok to say that `LICENSE` takes precedence over `COPYING`?

The logic here is meant to check this scenario.  It makes the assumption that rather than checking `[license file 1, license file 2, ... , package manager file, README]` one at a time as `licensed` was previously doing, it should be checking `[[all license files], package manager file, README]`.

Though based on ☝️ I should probably replace the existing `if` check with `project.license_files.map(&:license).uniq.size == 1`.

Does that make any more sense?  💭 ?

jonabc: > Does project.license there not evaluate licenses from license files, package manager files and readme files?

`project.license` from the if check there does, but `project.license_files` from the elsif check does not.

> I'm not sure when the code in 81-84 would obtain a different result, except where project.license returned other due to conflicting info in non-license files rather than multiple license files with conflicting info, but in that case other is still the right answer.

Yep!  That's exactly the case I was trying to describe.  I don't know that my logic is correct either in implementation or intentions but maybe an example would be better.

Looking at the tests, I think the relevant test cases that describe this logic are [finding licenses from non-license files](https://github.com/github/licensed/blob/a521ff925a77dc62e7366ac45461ec22cfd60c26/test/dependency_test.rb#L56-L75) and [how to handle license files that don't match](https://github.com/github/licensed/pull/64/files/4f04102e07388a2926d8362485aacd40641ee0d7#diff-af404304c2cc2c097e31ce95b4104687R142).  Are the outcomes correct for those scenarios?

lib/licensed/dependency.rb
```diff
@@ -68,24 +68,24 @@ def local_files
...
-    def matched_project_file
-      @matched_project_file ||= project.matched_files
-                                       .select { |f| f.license && !f.license.other? }
-                                       .first
+    # Finds and returns a license from the Licensee::FSProject for this dependency.
+    def project_license
+      @project_license ||= if project.license && !project.license.other?
+        # if Licensee has determined a project to have a specific license, use it!
+        project.license
+      elsif project.license_files.map(&:license).uniq.size > 1
```
benbalter: I _think_ this logic is [already encapsulated in `project.license`](https://github.com/benbalter/licensee/blob/master/lib/licensee/projects/project.rb#L22-L29)?

lib/licensed/dependency.rb
```diff
@@ -68,24 +68,24 @@ def local_files
...
+        # if there are multiple license types found from license files,
+        # return 'other'
+        Licensee::License.find("other")
+      else
+        # check other project files if we haven't yet found a license type
+        project.licenses.reject(&:other?).first
```
benbalter: If a project is dual licensed, it will pick the first? What if it's dual licensed as GPL and Apache, in that order? Perhaps we should look for a non-copyleft license first to avoid false positives when a project is dual licensed (e.g., "choose" to use Apache over GPL in that case).

lib/licensed/dependency.rb
jonabc: Yea @mlinksva and I were discussing that at https://github.com/github/licensed/pull/64#discussion_r201164703.

TL;DR it's not quite the same due to `project.license` including additional sources over `project.license_files`

lib/licensed/dependency.rb
jonabc: This was added to maintain legacy behavior where `licensed` previously chose the first non-other license found from `project.matched_files`.  If it's a good idea to intentionally change that behavior, 👍

lib/licensed/dependency.rb
mlinksva: I made the logic a bit more conservative in ac4b4ce. If there are any potentially conflicting license indicators, we get `other`, requiring human review. Not attached to implementation of course.

lib/licensed/dependency.rb
```diff
@@ -99,25 +89,16 @@ def remote_license_file
     # from the local LICENSE-type files, remote LICENSE, or the README, in that order
     def license_text
       content_files = Array(project.license_files)
-      content_files << remote_license_file if content_files.empty? && remote_license_file
+      content_files << remote_license_file if content_files.empty? && remote_license_file && remote_license_file.license.key == license_key
```
jonabc: if we're not using a remote file to find the license key, maybe we can 🔪 the remote file entirely?

mlinksva: Fine with me. As I wrote in the commit message, it is a minor convenience:

> leave one convenience: pull license text from github if no local
> text and license on github is the same as asserted in package manager

PR #65, opened 2018-07-10
Remove fetching of license content from GitHub

jonabc:
As discussed at https://github.com/github/licensed/pull/64#discussion_r201485343, we no longer are looking to pull remote license files from GitHub. 

This PR removes all code related to using `Licensed.from_github` to fetch remote files.


jonabc:
@mlinksva I think it would be good to remove this if we're not using the file to set the license key.  I can see the confusion around "the text is a perfect match to <insert license here>, why isn't the license key being found?"

jonabc:
For posterity, this is a non-breaking-change.

The `--offline` CLI option is the only user-facing change in this PR.  It's been deprecated but not removed, so that any existing tooling or scripting will continue to work

PR #67, opened 2018-07-11
Support Gemfile 'path' option for gem source.

vestedpr-dev:
Gems loaded via Gemfile `path:` option fail with `RuntimeError: Unable to find a specification for <gem> in any sources`

My app uses the CBRA ["Component Based Rails Applications" ](http://shageman.github.io/cbra.info/) methodology.  One of the key parts of this is, to divide the application into "components" which, are really just Rails Engines.  These engines are then loaded into a root, container app, via a Gemfile "path:" option, like so:

    gem 'api', '0.0.1', path: 'components/api'

When running `licensed` against this app, I get this error:

```
RuntimeError: Unable to find a specification for messaging (= 0.0.1) in any sources
  /.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/licensed-1.2.0/lib/licensed/source/bundler.rb:74:in `block in specs_for_dependencies'
  /.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/licensed-1.2.0/lib/licensed/source/bundler.rb:73:in `map'
```

Tracking it down, it only took me a few minutes to solve the problem, via changing this line in licensed's `bundler.rb` file:

```
- if spec.source.is_a?(::Bundler::Source::Gemspec)
+ if spec.source.is_a?(::Bundler::Source::Gemspec) || spec.source.is_a?(::Bundler::Source::Path)
  return spec.source.specs.first
end
```

If you think this change makes sense and, is harmless, than, perhaps this can be merged into the mainstream.

Thank you!

### DIFF
--- a/lib/licensed/source/bundler.rb
+++ b/lib/licensed/source/bundler.rb
@@ -90,7 +90,7 @@ module Licensed
 
         # if the specification is coming from a gemspec source,
         # we can get a non-lazy specification straight from the source
-        if spec.source.is_a?(::Bundler::Source::Gemspec)
+        if spec.source.is_a?(::Bundler::Source::Gemspec) || spec.source.is_a?(::Bundler::Source::Path)
           return spec.source.specs.first
         end
 
--- a/test/fixtures/bundler/Gemfile
+++ b/test/fixtures/bundler/Gemfile
@@ -22,3 +22,5 @@ end
 group :test do
  gem "minitest", "5.11.3"
 end
+
+gem "pathed-gem-fixture", path: "pathed-gem-fixture"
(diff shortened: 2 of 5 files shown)


jonabc:
@vestedpr-dev Thank you for opening this PR!  The change looks good, and your description is clear and makes sense!

Would you mind adding some tests to verify the changes in the PR?

jonabc:
@vestedpr-dev I was notified that this is showing up in other environments as well.  Since there hasn't been any additional activity, I'm going to push up some tests to demonstrate / verify the fix.

Thanks for your work in figuring out this out!

vestedpr-dev:
Thanks for that @jonabc.  Thank you for making the licensed gem.

Issue #71, opened 2018-07-24
Regression in 1.2.0


AlanFoster:
Hi, I noticed a regression between version 1.2.0 and 1.1.0 when trying to run `licensed status` on a project which has a dependency on `libv8`:

> /Users/User/.rvm/gems/ruby-2.5.1/gems/licensed-1.2.0/lib/licensed/source/bundler.rb:74:in `block in specs_for_dependencies': Unable to find a specification for libv8 (>= 6.3) in any sources (RuntimeError)

This error happens on both a mac + ubuntu with v1.2.0, but does _not_ happen on v1.1.0

----

### Replicating

With the following Gemfile:

```ruby
# frozen_string_literal: true

source "https://rubygems.org"

git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }

gem "mini_racer"
```

Gemfile.lock

```
GEM
  remote: https://rubygems.org/
  specs:
    libv8 (6.7.288.46.1)
    mini_racer (0.2.0)
      libv8 (>= 6.3)

PLATFORMS
  ruby

DEPENDENCIES
  mini_racer

BUNDLED WITH
   1.16.1
```

`.licensed.yml`

```
sources:
  rubygem: true
```

Running the following with licensed 1.2.0 currently fails:

```
licensed status
```

### Version 1.2.0

```
/Users/User/.rvm/gems/ruby-2.5.1/gems/licensed-1.2.0/lib/licensed/source/bundler.rb:74:in `block in specs_for_dependencies': Unable to find a specification for libv8 (>= 6.3) in any sources (RuntimeError)
```

### Version 1.1.0

```
licensed status
Checking licenses for example: 3 dependencies
FFF

Warnings:

/private/tmp/example/.licenses/rubygem/bundler.txt:
  - cached license data missing

/private/tmp/example/.licenses/rubygem/libv8.txt:
  - cached license data missing

/private/tmp/example/.licenses/rubygem/mini_racer.txt:
  - cached license data missing

3 dependencies checked, 3 warnings found.
```

### Extra debugging

I can see that `licensed` is trying to check the following `spec_path` locations:

```
/Users/user/.rvm/gems/ruby-2.5.1/specifications/mini_racer-0.2.0.gemspec
/Users/user/.rvm/gems/ruby-2.5.1/specifications/libv8-6.7.288.46.1.gemspec
```

It's dying currently dying on libv8, but I can tell that the `gemspec` actually has an extra operating system extension that `licensed` wasn't expecting:

```
[...]

PR #72, opened 2018-07-25
Find bundler specifications that include platform information

jonabc:
Fixes https://github.com/github/licensed/issues/71

The problem in the issue is that the installed gem is platform specific and includes `x86_64-darwin-16` in the gem file name.  The `LazySpecification` specs that licensed was using apparently wasn't platform aware, and `LazySpecification#full_name` didn't include the platform suffix.

It's possible to obtain a gem specification that is platform aware through `LazySpecification#source`.  I'm not sure that this is a 100% fix because bundler does a LOT internally, but it resolves the issue and doesn't seem to be a change for the worse.

This is a small change to try and find a spec in `spec.source.specs` that matches the lazy specifications name and version.  If found, it will be used to find the relevant `gemspec`.

### DIFF
--- a/lib/licensed/source/bundler.rb
+++ b/lib/licensed/source/bundler.rb
@@ -94,6 +94,13 @@ module Licensed
           return spec.source.specs.first
         end
 
+        # spec.source.specs gives access to specifications with more
+        # information than spec itself, including platform-specific gems.
+        # try to find a specification that matches `spec`
+        if source_spec = spec.source.specs.find { |s| s.name == spec.name && s.version == spec.version }
+          spec = source_spec
+        end
+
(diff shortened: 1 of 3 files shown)


jonabc:
> My bundler-fu also doesn't go as far as figuring out whether there would be other pieces/ways to check, but name + version should match the platform when specified, and I can't imagine a situation in which we'd match a bad candidate. 👍

👍 ! For posterity, it's also worth mentioning that licensed runs after everything has already been installed locally.  There is minimal confusion in terms of finding the right gems because everything is local, so things like version conflicts on the same gem and gem platform resolution will already be resolved.

Issue #77, opened 2018-09-06
Licensed doesn't traverse into submodules


lildude:
Whilst implementing using the manifest method for caching the license metadata in Linguist in https://github.com/github/linguist/pull/4267, I discovered that Licensed doesn't traverse submodules.

Digging into the code, it appears it is because of the call to `git ls-tree --full-tree -r --name-only HEAD` at https://github.com/github/licensed/blob/551d9b07440ca1287c20dae147cda902dfe1d26c/lib/licensed/git.rb#L38-L42 doesn't traverse submodules.

Switching this out for `git ls-files --recurse-submodules --full-name` is likely to get us the behaviour needed to traverse submodules.

PR #78, opened 2018-09-07
Use `git ls-files` instead of `git ls-tree` to find files in the git repo

jonabc:
Fixes https://github.com/github/licensed/issues/77

The manifest source is not finding files inside of submodules due to using `git ls-tree` to find files in the git repo.  This PR uses `git ls-files --recurse-submodules --full-name` instead which can enumerate files from within submodules.

Using `--full-name` returns the full path to the files from the repository root, no matter where in the repository the command is called.

In looking up the differences between the two functions online, it seems like `git ls-files` returns files based on the cached index while `git ls-tree` is finding files specifically for the `HEAD` tree.  The difference should be negligible though, as both solutions require the source to be run after the last commit to pick up the most recent commit hash for versioning.

### DIFF
--- a/test/license_test.rb
+++ b/test/license_test.rb
@@ -66,7 +66,7 @@ describe Licensed::License do
     end
 
     it "returns true if the normalized content is the same" do
-      license = Licensed::License.new({}, "test content")
+      license = Licensed::License.new({}, "- test content")
       other = Licensed::License.new({}, "* test content")
 
       assert license.license_text_match?(other)
--- a/lib/licensed/git.rb
+++ b/lib/licensed/git.rb
@@ -37,7 +37,7 @@ module Licensed
 
       def files
         return unless available?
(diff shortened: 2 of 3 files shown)


test/license_test.rb
```diff
@@ -66,7 +66,7 @@
     end
 
     it "returns true if the normalized content is the same" do
-      license = Licensed::License.new({}, "test content")
+      license = Licensed::License.new({}, "- test content")
```
jonabc: normalization rules changed in `licensee` that caused this to fail.  a leading `*` is now normalized as a list indicator rather than general markup, so I've updated the test to check that it's normalized to another list indicator `-`

jonabc:
I've built and tested `licensed` with this change manually for both manifest + submodule and go sources.  Everything looks good and I think with https://github.com/github/licensed/pull/78/commits/036a0dfdd6684cab629625d45d0c249434509bc2 tests should be 💚  .

/cc @lildude FYI this took a little longer to put together than I was expecting

PR #84, opened 2018-10-01
COPYING is a license file, shouldn't need to be treated as additional legal file

mlinksva:
In short, I don't think `Licensed::Dependency::LEGAL_FILES` needs to match anything that `Licensee::ProjectFiles::LicenseFile::FILENAME_REGEXES` matches because licensee will provide texts matching the latter.

I don't think this will cause any behavior change.

### DIFF
--- a/lib/licensed/dependency.rb
+++ b/lib/licensed/dependency.rb
@@ -3,7 +3,7 @@ require "licensee"
 
 module Licensed
   class Dependency < License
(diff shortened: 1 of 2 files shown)


saltvedt:
JRuby is a good edge case, I suppose:

https://github.com/jruby/jruby/blob/master/COPYING

mlinksva:
It's not really an edge case, there are lots of projects that include third party info in their `LICENSE`, `COPYING`, or similar files. Licensed will get the contents of those files from Licensee whether the latter can positively detect a license from them or not, thus Licensed doesn't need to collect them, thus this PR. 😄

Issue #1, opened 2018-02-02
Remove hard restriction on running in a Git repository


jonabc:
`licensed` currently requires a Git repository to translate relative configuration paths into absolute paths on the local machine. This is because

1. all current usages use git repositories
2. using absolute paths in a checked in configuration file is 👎
3. licensed requires absolute paths

The combination of ☝️ led to an implementation that only included support for Git.

For broader usage it would make sense to support `licesned` with non-Git source control and directories not in a source control system.

PR #94, opened 2018-10-14
Add option to configure root path, removing hard dependency on git

jonabc:
Fixes https://github.com/github/licensed/issues/1

This PR adds the ability to configure a `root` property that will be used to calculate paths used in `licensed`.  As the linked issue describes, using absolute paths in a configuration file isn't a good idea and as a result all configured paths are expected to be resolved based on a root.

Currently there is a hard dependency on git being available to use licensed because the git repository root is used as the root for all path expansion.  This change removes that hard dependency by allowing users to manually configure a root path using the configuration `root` property.

If a configuration file specifies a path for `root`, the relative path will be expanded to an absolute path after parsing the file contents.  For example, the following configuration sets a root path relative to the configuration file directory.

```yaml
# configuration file at ~/licensed/test/project/.licensed.yaml

root: "../.."                                                    # ~/licensed
source_path: "test/project/subproject"    # ~/licensed/test/project/subproject
```

Alternatively, a configuration file can set `root` to the boolean `true` to use the configuration file's directory as the root path.  For the example above, setting `true` would look like:

```yaml
# configuration file at ~/licensed/test/project/.licensed.yaml

root: true                                   # ~/licensed/test/project
source_path: "subproject"       # ~/licensed/test/project/subproject
```

Root path expansion will only occur when set in a configuration file.  If a `root` property is passed to `Licensed::Configuration.new` it will not be adjusted and is expected to be an absolute path.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -4,6 +4,16 @@ A configuration file specifies the details of enumerating and operating on licen
 
 Configuration can be specified in either YML or JSON formats.  Examples below are given in YML.
 
+## Configuration Paths
+
+`licensed` requires a path to enumerate dependencies at (`source_path`) and a path to store cached metadata (`cache_path`).
+
+To determine these paths across multiple environments where absolute paths will differ, a known root path is needed to evaluate relative paths against.
+In using a single root, relative source and cache paths can be specified in the configuration file.
+
+A root path can be manually configured as a path relative to the configuration file using the `root` property.  If a root path is not specified,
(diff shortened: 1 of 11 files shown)


jonabc:
One note on this change.  licensed doesn't really handle the situation when a root isn't available and will end up raising an error on trying to call `.join` on the nil `AppConfiguration#root`.

This problem existed before this PR when licensed was called outside of a git repository.  I'm going to open an issue to handle a nil root separately from this PR.

Issue #98, opened 2018-10-23
Ignoring NPM optionalDependencies


mwagz:
I'm currently using `licensed` in my project. This project also has a dependency of the Webpacker gem, which includes an npm dependency of `@rails/webpacker`. `fsevents` is an optional dependency through the following dependency tree:

```
@rails/webpacker => webpack => watchpack => chokidar => fsevents
```

This means that `licensed status` will detect it. The problem is that `fsevents` is an `optionalDependency` in this particular NPM package. So if it fails to install, the package just falls back to IIRC, but `licensed` still wants it to be present. So I receive the output in [this Gist](https://gist.github.com/mwagz/eb758a261619ce05bb453dcf2e7a0b37).

I think, ideally, npm packages that are "optional" would be ignored by licensed. Adding it to ignored does not seem to work.

jonabc:
@mwagz thanks for the bug report.  If you want to open a PR to fix this issue I'd be happy to review it, otherwise I'll try to get to this soon.

mwagz:
@jonabc Thanks and sure. I'm looking at it now. It seems like we're failing because it can't find it in `node_modules` before we check if it's ignored. I'm going to try to check if it's be designated as ignored before we fail. I'll let you know if I'm unable to figure anything out.

PR #100, opened 2018-10-23
Stop Validation on Ignored NPM Packages

mwagz:
Fixes: https://github.com/github/licensed/issues/98

An NPM package could be ignored, but the NPM source would fail if it was unable to find the package within the node_modules directory. If we are ignoring the package, then we should not be concerned with that package's existence in node_modules.

Skip over the assets in building the dependencies list if the package is ignored.


mwagz:
@jonabc I'm relatively new to `minitest` and only moderately up to speed with Ruby. I think the test is doing what I expect it to, but if there's a better way to omit these things, I'm all about learning it!

Thanks!

lib/licensed/source/npm.rb
```diff
@@ -34,6 +34,7 @@ def dependencies
       def packages
         root_dependencies = JSON.parse(package_metadata_command)["dependencies"]
         recursive_dependencies(root_dependencies).each_with_object({}) do |(name, results), hsh|
+          next if @config.ignored?(hsh)
```
jonabc: I don't think this will work the way you're expecting.  `hsh` in this case is an accumulator of `{ <name> => { <package data> } }` pairs, and doesn't have `@config.ignored?`s expected structure of `{ "name" => <name>, "type" => <type> }`

test/source/npm_test.rb
```diff
@@ -54,6 +54,13 @@
         end
       end
 
+      it "does not include ignored dependencies" do
+        Dir.chdir fixtures do
+          @config.ignore({ type: "npm", name: "autoprefixer" })
```
jonabc: `@config.ignore` is expecting string keys `"type" => Licensed::Source::NPM.type, "name" => "autoprefixer"`

lib/licensed/source/npm.rb
mwagz: Ahh that makes sense! Thanks for the insight there. I was able to structure this in the same manner as you outlined for Go and it seems to be excluding as I expect it to. The changes I made in [Licensed::Command::List](https://github.com/github/licensed/blob/24b60232243f5773a16318ab51fe5f713e25c81e/lib/licensed/command/list.rb) did not seem to be necessary.

mwagz:
@jonabc Thanks for pointing me in the right direction. I've made the updates to be more closely aligned to the logic you outlined. I _think_ the test is doing what I expect it to. I was able to verify (via `puts` in the Source module) that `autoprefixer` is being omitted for that test.

Thanks again!

mwagz:
@jonabc Would you be willing to cut a release with this commit in it? No big deal if not, I can have the gem point to my fork until the next release, but thought I'd ask. Thanks!

jonabc:
@mwagz yep!  I have a few other changes coming in very soon and was going to cut a release with all of them soon 👍

jonabc:
@mwagz https://github.com/github/licensed/releases/tag/1.5.0 🎉

mwagz:
@jonabc 🙌 You just helped me close a 3 month old PR! Thanks man!

Issue #62, opened 2018-06-27
Add a version command


sridharavinash:
Now that we have a binary build with `ruby-packer` it would be nice to have a 
```
licensed --version
```
which would display the current version of the binary.
This could be useful for applications that want to download and run the `licensed` binary and automatically update to the latest version if available.
 for .eg
```
licensed --update-licensed
```
would update the current copy is there's a later release version of the binary otherwise continue using the locally cached version.

Does that make sense?

dideler:
I believe this can be closed now.

```
$ licensed version
1.5.2

$ licensed --version
1.5.2
```

jonabc:
@sridharavinash I'm closing this issue (agree with ☝️ ).

> `licensed --update-licensed`

It will be hard to scope down which update the user wants based on the scenario of usage.  Does the user want the latest patch update?  minor release? etc.

It might also cause headaches if not used properly and the pulled update contains breaking changes.

I think this is better left to manual interaction or bespoke automation built for specific scenarios.  If you still think this is worth investigating, please open a separate issue 🙇

PR #101, opened 2018-10-24
Add Version Command

mwagz:
Addresses 1 part of https://github.com/github/licensed/issues/62. I'm not entirely certain if the issue is suggesting that licensed is able to update itself with the second part, or if he was just throwing around some ideas. The `version` command would at least allow users to automatically update based on the output of `version`.

**Usage**
```shell
licensed version
=> 1.4.0
licensed -v
=> 1.4.0
licensed --version
=> 1.4.0

### CODE REVIEW

File: lib/licensed/command/version.rb [hunk 1]
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+module Licensed
+  module Command
+    class Version
+      def initialize
+      end
+
+      def run
+        $stdout << "#{Licensed::VERSION}\n"
```
jonabc: It's a bit of overkill for whats needed here, but could you use `Licensed::UI::Shell#info` to output the version?  That would keep output methods consistent throughout the program.  🙇

mwagz: Makes sense. I was a little back and forth between `$stdout` and `puts`. `Licensed::UI::Shell#info` answers that question. 😂

mwagz:
Switched from `$stdout` to `Licensed::UI::Shell#info`. Thanks!

PR #102, opened 2018-10-24
Use a safer check for if a directory is a git repo

dbussink:
This uses rev-parse to check whether a repository is a git repository. This is a safer operation than using git status, as git status might end up modifying .git/index and updating it, especially in the case of a dirty tree.

### DIFF
--- a/lib/licensed/git.rb
+++ b/lib/licensed/git.rb
@@ -16,7 +16,7 @@ module Licensed
 
       # Returns true if a git repository is found, false otherwise
       def git_repo?
-        repository_root.empty?
+        !repository_root.empty?
       end
 
       # Returns the most recent git SHA for a file or directory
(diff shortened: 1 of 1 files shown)


lib/licensed/git.rb
```diff
@@ -9,7 +9,7 @@ def available?
 
       def git_repo?
         return false unless available?
-        !Licensed::Shell.execute("git", "status", allow_failure: true).empty?
+        !Licensed::Shell.execute("git", "rev-parse", "HEAD", allow_failure: true).empty?
```
jonabc: 🤔 `repository_root` already uses `git rev-parse --show-toplevel`, is that heavier weight than `git rev-parse HEAD`?  maybe this could be reworked like

```
def repository_root
  return unless available?
  Licensed::Shell.execute("git", "rev-parse", "--show-toplevel", allow_failure: true)
end

def git_repo?
  repository_root.empty?
end
```

dbussink: @jonabc That might be possible yeah, you're probably the better expert here than me. If that works as well, might be a good further simplification.

Issue #99, opened 2018-10-23
Packaged executable ruby version mismatch causes bundler source to fail


jonabc:
Licensed distributes all-in-one executables to make running licensed outside of a ruby environment possible.  The executables are created using [`ruby-packer`](https://github.com/pmq20/ruby-packer) and builds an executable from a squashed filesystem containing ruby and licensed.

ruby-packer is currently using ruby version 2.4.X, which causes `Bundler.specs_path` to always evaluate to `<gem path root>/ruby/2.4.0` - in my usages this was `vendor/gems/ruby/2.4.0`.

Licensed uses [`::Bundler.specs_path`](https://github.com/github/licensed/blob/24b60232243f5773a16318ab51fe5f713e25c81e/lib/licensed/source/bundler.rb#L99) to find specification files and load full information for each dependency.

When running `bundle install` with a ruby version that's not 2.4.x, specifications will not be under `<gem path root>/ruby/2.4.0` and licensed will raise an error that specifications could not be found.

In the below example, `jekyll` was `bundle install`'d using ruby 2.5.0 and was present under `vendor/gems/ruby/2.5.0`.

```
/__enclose_io_memfs__/lib/ruby/gems/2.4.0/gems/licensed-1.3.0/lib/licensed/source/bundler.rb:74:in `block in specs_for_dependencies': Unable to find a specification for jekyll (>= 0) in any sources (RuntimeError)
```

PR #106, opened 2018-10-29
Use ruby version from host for enumerating gems with licensed executable

jonabc:
Fixes https://github.com/github/licensed/issues/99

As mentioned in the attached issue, there is the potential for a ruby version mismatch when using the licensed executable.  This PR solves the problem by obtaining and setting `Gem::ConfigMap[:ruby_version]` from the host environment when running licensed from the built executable.  `Gem::ConfigMap[:ruby_version]` is used by bundler to construct it's gem and specification paths.

When the ruby version is set to the `Gem::ConfigMap` value found in the host environment, bundler can construct paths that match the host ruby environment rather than the ruby environment packaged into the licensed executable.

### DIFF
--- a/docs/sources/bundler.md
+++ b/docs/sources/bundler.md
@@ -2,6 +2,18 @@
 
 The bundler source will detect dependencies `Gemfile` and `Gemfile.lock` files are found at an apps `source_path`.  The source uses the `Bundler` API to enumerate dependencies from `Gemfile` and `Gemfile.lock`.
 
+### Enumerating bundler dependencies when using the licensed executable
+
+**Note** this content only applies to running licensed from an executable.  It does not apply when using licensed as a gem.
+
+_It is strongly recommended that ruby and bundler are always available when running the licensed executable._
+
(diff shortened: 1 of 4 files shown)


jonabc:
@mwagz nope!  It only covers the scenario when the licensed executables which are attached to each [GitHub release](https://github.com/github/licensed/releases/tag/1.5.0) are built using, for example, ruby 2.4.2 but you use ruby 2.5.0 in your local environment.

`bundle install` will end up installing gems and specifications to a path like `.../ruby/2.5.0/...`, but the licensed executable will be looking at `.../ruby/2.4.0/...`

jonabc:
One additional note for posterity before merging:

This PR doesn't enforce that ruby and bundler are available when running licensed executables, however it is strongly encouraged.  There are edge cases that can occur when the ruby version used when `bundle install` is run isn't known.

I went back and forth on this a little because if you have ruby available, then why not `gem install` or `bundle install` the gem directly?  Ultimately I think this is an ok tradeoff because it allows a single workflow for running licensed in a containerized, CI environment without having to choose between different install paths for different dependency types.

PR #107, opened 2018-10-30
Add description how to write setting of pip and python

kusuwada:
Because I can not find where and how to set the settings for pip & python, I worried for a while and read source code. So I added it to the README.
I believe this will be of help to those who use this library.

### DIFF
--- a/docs/sources/pip.md
+++ b/docs/sources/pip.md
@@ -19,9 +19,6 @@ The `pip` command will be sourced from this directory.
 You have to add this setting to `.licensed.yaml`.
 An example usage of this might look like:
 ```yaml
-sources:
-  pip: true
-
 python:
     virtual_env_dir:"/path/to/your/venv_dir"
 ```
--- a/docs/sources/pip.md
+++ b/docs/sources/pip.md
@@ -16,7 +16,7 @@ _note_: `<your_venv_dir>` path should be relative to the repository root or can 
 #### virtual_env_dir (Required)
 
 The `pip` command will be sourced from this directory.
-You have to add this setting to `.licensed.yaml`.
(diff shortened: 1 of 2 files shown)


jonabc:
@kusuwada thanks for the contribution!  The changes you've described apply to all sources, not just pip and python, and is documented in [`docs/configuration.md`](https://github.com/github/licensed/blob/master/docs/configuration.md#restricting-sources).

To be clear, it is usually not required to explicitly enable any sources in the licensed configuration file.  Setting a source to `true` as you've suggested is only needed when you've explicitly set another source to true.

If that documentation is unclear, or it's not easy to find, I would very happily accept any updates that  improve that documentation, but I would prefer not to merge this PR as-is.

kusuwada:
@jonabc OK, I'll try to update documents. Please wait for a while.

kusuwada:
It's not explicitly specifyed what to specify for `source` to README, and I think it was little confusion for user, so I add these.
Also, it was written in the README to execute all sources with default, but I wrote more clearly.

docs/sources/pip.md
```diff
@@ -16,6 +16,7 @@ _note_: `<your_venv_dir>` path should be relative to the repository root or can
 #### virtual_env_dir (Required)
 
 The `pip` command will be sourced from this directory.
+You have to add this setting to `.licensed.yaml`.
```
jonabc:
```suggestion
You have to add this setting to your licensed configuration file.
```

Users might be using a different configuration file name

kusuwada: That's right. I fix this as it.

jonabc:
Thanks for the docs update @kusuwada 🙇 !

PR #113, opened 2018-12-27
Add support for golang 1.11+, go modules

jonabc:
This PR updates the Go dependency source enumerator to better support newer versions of the Go language.

1. Uses the `go list -deps` option to get all dependencies in a single call rather than needing to call `go list` for each dependency.
2. If a go project is set up to use modules, uses the modules version number instead of a git SHA1 for version information

This ended up including a non-trivial refactor of the go dependency source to give an easy branching point for sourcing package information for older go versions vs `go list -deps` as described in (1) ☝️ .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ test/fixtures/npm/package-lock.json
 test/fixtures/go/src/*
 test/fixtures/go/pkg
 !test/fixtures/go/src/test
+!test/fixtures/go/src/modules_test
 test/fixtures/cabal/*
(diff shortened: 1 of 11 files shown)


jonabc:
> Uses the go list -deps option to get all dependencies in a single call rather than needing to call go list for each dependency.

In the CI environment, the go 1.11.x tests consistently run ~1minute, ~33% faster than for previous versions due to this speedup 🎉 !  And thats only for 2-3 dependencies in the test project.  For go projects with a significant number of dependencies the change will be even more dramatic.

PR #134, opened 2019-02-03
Print 'license needs review' when license is not allowed

danielcompton:
Currently licensed prints like this when a license is not allowed:

```
.licenses/rubygem/bundler.txt:
  - license needs reviewed: mit.
```

It's a very minor detail, but I think it might read better to say "license needs review: mit."?


jonabc:
Looks great!  Thanks for the update @danielcompton !

PR #151, opened 2019-02-26
Find gradle dependency licenses via new dependency subclass

jonabc:
:wave: @dbussink here's what I had in mind about loading dependency license contents using a subclass.  I don't have gradle locally so haven't yet run any tests to see if this actually works, but it hopefully is enough to demonstrate the idea.

### DIFF
--- a/lib/licensed/dependency.rb
+++ b/lib/licensed/dependency.rb
@@ -74,47 +74,45 @@ module Licensed
     def license_contents
       matched_files.reject { |f| f == package_file }
                    .group_by(&:content)
-                   .map { |content, files| { "sources" => content_sources(files), "text" => content } }
+                   .map { |content, files| { "sources" => license_content_sources(files), "text" => content } }
     end
 
     # Returns legal notices found at the dependency path
     def notice_contents
-      notice_files.sort # sorted by the path
-                  .map { |file| { "sources" => content_sources(file), "text" => File.read(file).rstrip } }
-                  .select { |text| text.length > 0 } # files with content only
-    end
-
-    # Returns an array of file paths used to locate legal notices
-    def notice_files
       Dir.glob(dir_path.join("*"))
          .grep(LEGAL_FILES_PATTERN)
          .select { |path| File.file?(path) }
+         .sort # sorted by the path
+         .map { |path| { "sources" => normalize_source_path(path), "text" => File.read(path).rstrip } }
+         .select { |notice| notice["text"].length > 0 } # files with content only
     end
 
     private
 
-    # Returns the sources for a group of license or notice file contents
+    # Returns the sources for a group of license file contents
     #
     # Sources are returned as a single string with sources separated by ", "
-    def content_sources(files)
+    def license_content_sources(files)
       paths = Array(files).map do |file|
-        path = if file.is_a?(Licensee::ProjectFiles::ProjectFile)
-          dir_path.join(file[:dir], file[:name])
-        else
-          Pathname.new(file).expand_path(dir_path)
-        end
+        next file[:uri] if file[:uri]
 
-        if path.fnmatch?(dir_path.join("**").to_path)
-          # files under the dependency path return the relative path to the file
-          path.relative_path_from(dir_path).to_path
-        else
-          # otherwise return the source_path as the immediate parent folder name
-          # joined with the file name
-          path.dirname.basename.join(path.basename).to_path
-        end
+        path = dir_path.join(file[:dir], file[:name])
+        normalize_source_path(path)
       end
 
       paths.join(", ")
+    end
+
+    def normalize_source_path(path)
+      path = Pathname.new(path) unless path.is_a?(Pathname)
+      if path.fnmatch?(dir_path.join("**").to_path)
+        # files under the dependency path return the relative path to the file
+        path.relative_path_from(dir_path).to_path
+      else
+        # otherwise return the source_path as the immediate parent folder name
+        # joined with the file name
+        path.dirname.basename.join(path.basename).to_path
+      end
     end
 
     # Returns the metadata that represents this dependency.  This metadata
--- a/lib/licensed/sources/gradle.rb
+++ b/lib/licensed/sources/gradle.rb
@@ -43,12 +43,6 @@ module Licensed
         end
       end
 
-      def with_latest_licenses
-        download_and_cache_licenses do
-          yield
-        end
-      end
-
       private
 
       def download_licenses_from_csv
--- a/lib/licensed/sources/gradle.rb
+++ b/lib/licensed/sources/gradle.rb
@@ -9,9 +9,9 @@ module Licensed
   module Sources
     class Gradle < Source
 
-      DEFAULT_CONFIGURATIONS   = ["runtime", "runtimeClasspath"]
-      GRADLE_LICENSES_PATH     = ".gradle-licenses"
-      GRADLE_LICENSES_CSV_NAME = "licenses.csv"
+      DEFAULT_CONFIGURATIONS   = ["runtime", "runtimeClasspath"].freeze
+      GRADLE_LICENSES_PATH     = ".gradle-licenses".freeze
+      GRADLE_LICENSES_CSV_NAME = "licenses.csv".freeze
 
       class Dependency < Licensed::Dependency
         class << self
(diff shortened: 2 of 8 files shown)


lib/licensed/sources/gradle.rb
```diff
@@ -9,9 +9,31 @@ module Licensed
...
+        def exist?
+          # shouldn't force network connections just to check if content exists
+          # only check that the path is not empty
+          !path.to_s.empty?
+        end
+
+        # Returns a Licensee::ProjectFiles::LicenseFile for the dependency
+        def project_files
+          return [] if path.to_s.empty?
+          Array(Licensee::ProjectFiles::LicenseFile.new(self.class.license(path), path))
```
jonabc: something missing here is error handling.  ideally any possible errors raised by `Net::HTTP.get(uri)` would be rescued, an error message would be set to `errors` and then added to the report in `Licensed::Commands::Cache.evaluate_dependency`

lib/licensed/sources/gradle.rb
```diff
@@ -98,38 +98,38 @@ def executable
...
-            configurations[it].resolvedConfiguration.resolvedArtifacts.each { artifact ->
-                def id = artifact.moduleVersion.id
-                dependencies << "  { \\"group\\": \\"${id.group}\\", \\"name\\": \\"${id.name}\\", \\"version\\": \\"${id.version}\\" }"
-            }
-        }
-        println "[\\n${dependencies.join(", ")}\\n]"
-    }
-}
-EOF
+        <<~EOF
```
jonabc: squiggly heredoc to keep the contents indented to the method

lib/licensed/sources/gradle.rb
```diff
@@ -30,55 +52,33 @@ def enabled?
...
-        end
-      end
-
       private
 
-      def download_licenses_from_csv
+      def packages
+        metadata = JSON.parse(gradle_command("printDependencies"))
+        gradle_command("generateLicenseReport")
         path = File.join(@config.pwd, GRADLE_LICENSES_PATH)
```
jonabc: ah, I think I also missed removing the licenses path and CSV file after evaluating the CSV contents

dbussink:
> This takes the approach that generating the CSV isn't too costly and can be created on every run of this source. Is that idea valid?

This is something I have been explicitly trying to avoid, as generating the CSV adds the additional dependency listed in the Gradle file and also therefore implies potential network operations. The idea of the original change was that license verify could run without needing to install / run that additional dependency, so that's why I only added it to the cache step.

jonabc:
> This is something I have been explicitly trying to avoid, as generating the CSV adds the additional dependency listed in the Gradle file and also therefore implies potential network operations.

@dbussink 👍 thanks for the additional context!  I changed the code here so that the csv is loaded from in the dependency class.  It takes a little rework to make everything callable but otherwise I think this _should_ work while keeping all the changes needed localized to the gradle source.

Does this make more sense?  Is it something you can work with to completion?

lib/licensed/sources/gradle.rb
```diff
@@ -9,127 +9,152 @@ module Licensed
...
+          f.write gradle_file(configurations)
           f.close
           Licensed::Shell.execute(executable, "-q", "-b", f.path, *args)
         end
       end
 
-      def executable
-        if File.executable?(File.join(@config.pwd, "gradlew"))
+      def self.executable
+        if File.executable?(File.join(config.pwd, "gradlew"))
```
jonabc: ah, missed this one.  this would need to be cleaned up with either the path being passed in, or some other way to uniquely find the `gradlew` file

lib/licensed/sources/gradle.rb
```diff
@@ -9,127 +9,156 @@ module Licensed
...
+            DEFAULT_CONFIGURATIONS
+          end
+        end
       end
 
-      def gradle_command(*args)
-        Tempfile.open(["license", ".gradle"], @config.pwd) do |f|
-          f.write gradle_file
+      def self.gradle_command(*args, executable:, configurations:)
+        Tempfile.create do |f|
```
jonabc: I made a (possibly wrong) assumption that the location of the temporary file doesn't matter, and the outputs of the gradle command would be in the local directory.  If that's not the case, then the gradle command needs to be given a path that would be used instead of replace `@config.pwd`.

Also not sure why I changed open to create, probably just miming something I read from the web.  The end result should be the same I think.

lib/licensed/sources/gradle.rb
```diff
@@ -9,127 +9,156 @@ module Licensed
...
+
+        # Returns whether the dependency content exists
+        def exist?
+          # shouldn't force network connections just to check if content exists
+          # only check that the path is not empty
+          !path.to_s.empty?
+        end
+
+        # Returns a Licensee::ProjectFiles::LicenseFile for the dependency
+        def project_files
```
dbussink: @jonabc Hmm, I'm not sure how this works? I see `lib/licensed/sources/manifest.rb` also contains this method but I don't see it called anywhere? Am I missing something here?

jonabc: @dbussink with licensed v2, `Licensed::Dependency` inherits from `Licensee::FSProject` to be able to directly provide additional content to licensee for cases such as this or the manifest source.

One way to provide additional content is by overriding this method.  The base method is  [`Licensee::Projects::Project#project_files`](https://github.com/licensee/licensee/blob/da21bba0352f8086920ec59e61c8dbd93f4f0e6d/lib/licensee/projects/project.rb#L148), which is used internal to licensee as the source of license file content to evaluate.

jonabc:
> The CSV loading has the same issue that it also connects to the internet afaik and will end up downloading the plugin on demand as well. If the aim is to have license validation work without that as a dependency, I don't think this works?

It _should_, though as I said it's untested.

Finding dependency license files and evaluating their contents using `licensee` happens as a result of loading the [`Dependency#record`](https://github.com/github/licensed/blob/3780ceb25e2b964c5a9bed826e1a6f3beb819744/lib/licensed/dependency.rb#L61) data.  Similarly the `license` metadata field is set on-demand when evaluating the [metadata](https://github.com/github/licensed/blob/3780ceb25e2b964c5a9bed826e1a6f3beb819744/lib/licensed/dependency.rb#L131) for a dependency's record data.

So the solution in this PR hooks into the mechanism that is used to find license files via a method override from `Licensee::Projects::Project`, and that mechanism should only be triggered from the `cache` command.

dbussink:
@jonabc I pushed up some fixes to make this PR work, do the changes make sense to you?

jonabc:
@dbussink thanks for getting things working!  I made some followup changes, mostly cleanup and trying to reduce the amount of path data that needed to be considered.  I'll leave comments inline

lib/licensed/dependency.rb
```diff
@@ -74,49 +74,47 @@ def license_key
...
-
-        if path.fnmatch?(dir_path.join("**").to_path)
-          # files under the dependency path return the relative path to the file
-          path.relative_path_from(dir_path).to_path
-        else
-          # otherwise return the source_path as the immediate parent folder name
-          # joined with the file name
-          path.dirname.basename.join(path.basename).to_path
-        end
+        next file[:uri] if file[:uri]
```
jonabc: all of the changes in this file are to cleanly support this line which allows a `uri` property instead of needing a `file[:dir]` and `file[:name]` 😆

in this case `file[:uri]` would be used to express that the data was pulled from a location that shouldn't go through any path normalization

lib/licensed/sources/gradle.rb
```diff
@@ -4,132 +4,164 @@
...
+            @csv ||= begin
+              gradle_licenses_dir = File.join(path, GRADLE_LICENSES_PATH)
+              Licensed::Sources::Gradle.gradle_command("generateLicenseReport", path: path, executable: executable, configurations: configurations)
+              CSV.foreach(File.join(gradle_licenses_dir, GRADLE_LICENSES_CSV_NAME), headers: true).each_with_object({}) do |row, hsh|
+                name, _, version = row["artifact"].rpartition(":")
+                key = csv_key(name: name, version: version)
+                hsh[key] = row["moduleLicenseUrl"]
+              end
+            ensure
+              FileUtils.rm_rf(gradle_licenses_dir)
```
jonabc: I noticed the grade licenses dir was hanging around after running tests 👍

lib/licensed/sources/gradle.rb
```diff
@@ -4,132 +4,164 @@
...
+          # Loads and caches license report CSV data as a hash of :name-:version => :url pairs
+          #
+          # executable     - The gradle executable to run to generate the license report
+          # configurations - The gradle configurations to generate license report for
+          #
+          # Returns a hash of dependency identifiers to their license content URL
+          def load_csv(path, executable, configurations)
+            @csv ||= begin
+              gradle_licenses_dir = File.join(path, GRADLE_LICENSES_PATH)
+              Licensed::Sources::Gradle.gradle_command("generateLicenseReport", path: path, executable: executable, configurations: configurations)
```
jonabc: I removed `path.parent` here since it's much easier to reason about known path hierarchy requirements under the current directory than above and around it.  The dependency is passed `path: config.pwd` now instead of `gradle_licenses_dir` to account for this, with the `grade_licenses_dir` path being constructed (and deleted) in this method

lib/licensed/sources/gradle.rb
```diff
@@ -4,132 +4,164 @@
...
+          # Cache and return the results of getting the license content.
+          def retrieve_license(url)
+            (@licenses ||= {})[url] ||= Net::HTTP.get(URI(url))
+          end
+        end
+
+        def initialize(name:, version:, path:, executable:, configurations:, metadata: {})
+          @configurations = configurations
+          @executable = executable
+          super(name: name, version: version, path: path, metadata: metadata)
```
jonabc: removed `@path` since it's already tracked in `Licensed::Dependency#path`

lib/licensed/sources/gradle.rb
```diff
@@ -4,132 +4,164 @@
...
+          !path.to_s.empty?
+        end
+
+        # Returns a Licensee::ProjectFiles::LicenseFile for the dependency
+        def project_files
+          self.class.load_csv(path, @executable, @configurations)
+
+          url = self.class.url_for(self)
+          license_data = self.class.retrieve_license(url)
+          Array(Licensee::ProjectFiles::LicenseFile.new(license_data, { uri: url }))
```
jonabc: with the new `uri` property that's picked up in `Licensed::Dependency` there's no need to write a file to disk with the license data 👍

lib/licensed/sources/gradle.rb
```diff
@@ -4,132 +4,164 @@
...
       end
 
-      def executable
-        if File.executable?(File.join(@config.pwd, "gradlew"))
-          "./gradlew"
-        else
-          "gradle"
+      def self.gradle_command(*args, path:, executable:, configurations:)
+        Dir.chdir(path) do
+          Tempfile.create(["license-", ".gradle"], path) do |f|
```
jonabc: switched this back to `.create` after noticing that `.open` doesn't immediately remove the created file.  From the [docs](https://ruby-doc.org/stdlib-2.3.0/libdoc/tempfile/rdoc/Tempfile.html), it seems like this is the major difference between `.create` and `.open` when a block is given 🤷‍♂️

jonabc:
added gradle tests to CI 👍 but had to remove the test to verify behavior when gradle/gradlew aren't available.  I wasn't able to figure out how to get CI to not find `gradle` while finding everything else needed to run `licensed`. 🤷‍♂️ doesn't look like any of the other sources make much effort to verify behavior when the required language tools aren't available

jonabc:
@dbussink do you have any more thoughts here?  If you don't see any complications with this approach I think it's preferable to introducing the shared method that needs to be considered when writing or updating commands.

PR #157, opened 2019-04-04
Relax thor dependency

eileencodes:
I used < 2.0 because that's what Rails uses. Happy to make the high end more strict as long as we can support 0.20+. Thanks!

---

Rails 6.0 Railties requires thor set to
`s.add_dependency "thor", ">= 0.20.3", "< 2.0"` and currently licensed
won't allow anything higher than 0.19.x. See
https://github.com/rails/rails/blob/master/railties/railties.gemspec#L40

This PR relaxes the thor dependency so that licensed can be used with
applications running or upgrading to Rails 6+.

cc/ @jonabc

### DIFF
--- a/licensed.gemspec
+++ b/licensed.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = ">= 2.3.0"
 
   spec.add_dependency "licensee", "~> 9.10"
-  spec.add_dependency "thor", "~> 0.19"
+  spec.add_dependency "thor", "~> 0.19", "< 2.0"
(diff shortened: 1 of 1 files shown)


jonabc:
@eileencodes sorry for the holdup, the CI failure is from an unrelated change to rubocop that rubocop-github hasn't yet updated for.  I've just merged https://github.com/github/licensed/pull/158, can you merge master into your fork?  that should get CI passing again

eileencodes:
Hey it turns out I misread the output from Bundler (who hasn't?) and we don't actually need this. Sorry for the noise :smile:

PR #159, opened 2019-04-09
Handle plugins block and empty url in gradle source

jandersson-svt:
Most more advanced build.gradle files out in the wild has a plugins block,
buildscript block or both. At the moment there is a restriction
where plugins blocks can be added, so the current implementation
cant handle the "apply build.gradle" clause when generating the
tempfile.

This fix will check if the "Gradle license report" plugin is installed,
and if not, either add it to the plugins block, or add an
plugins block with it if there was none in the build.gradle file.
This should make the gradle source compatible with all the cases first mentioned.

Also, there was a problem where if one dependency was missing an url,
the license cache check would fail with exception.
This takes an easy way out to fix this - it sets it to an empty
mutable string.


jandersson-svt:
I tested the fix on a few real life projects.

lib/licensed/sources/gradle.rb
```diff
@@ -101,27 +108,50 @@ def enumerate_dependencies
...
+
+      def self.should_add_gradle_license_report_plugins_block?(gradle_build_file)
+        !/dependency-license-report/.match(gradle_build_file)
+      end
+
+      def self.build_gradle_contains_plugins_block?(gradle_build_file)
+        /plugins/.match(gradle_build_file)
+      end
+
+      def self.add_gradle_license_report_plugins_block?(gradle_build_file)
```
jonabc: if this isn't returning a boolean, it shouldn't end in `?` 👍

lib/licensed/sources/gradle.rb
```diff
@@ -101,27 +108,50 @@ def enumerate_dependencies
...
+        if(Licensed::Sources::Gradle.build_gradle_contains_plugins_block?(gradle_build_file))
```
jonabc: can you inline the check that occurs in the called method to make the code a bit simpler?
```suggestion
        if gradle_build_file.include?("plugins")
```

lib/licensed/sources/gradle.rb
```diff
@@ -101,27 +108,50 @@ def enumerate_dependencies
...
+          gradle_build_file.gsub(/(?<=plugins)\s+{/, " { id 'com.github.jk1.dependency-license-report' version '1.6'")
+        else
+
+          gradle_build_file = " plugins { id 'com.github.jk1.dependency-license-report' version '1.6' }" + gradle_build_file
         end
       end
 
       def self.gradle_command(*args, path:, executable:, configurations:)
+        gradle_build_file = File.read("build.gradle")
+        if(Licensed::Sources::Gradle.should_add_gradle_license_report_plugins_block?(gradle_build_file))
```
jonabc: the new code is not as easy to read as it could be due to long and similar looking method names.  what do you think about inlining this check into `add_gradle_license_report_plugins_block` as a guard clause

```
def self.add_gradle_license_report_plugins_block(gradle_build_file)
   return unless gradle_build_file.include? "dependency-license-report"
   ...
```

then all the logic to ensure that the gradle file is 👍 is self contained in the new method

lib/licensed/sources/gradle.rb
```diff
@@ -101,27 +108,50 @@ def enumerate_dependencies
       def gradle_executable
         return @gradle_executable if defined?(@gradle_executable)
         @gradle_executable = begin
-          gradlew = File.join(config.pwd, "gradlew")
-          return gradlew if File.executable?(gradlew)
-          "gradle" if Licensed::Shell.tool_available?("gradle")
-        end
+                               gradlew = File.join(config.pwd, "gradlew")
```
jonabc: same here, it doesn't look like anythings actually changed.  can you revert this block?

lib/licensed/sources/gradle.rb
```diff
@@ -101,27 +108,50 @@ def enumerate_dependencies
...
       # Returns the configurations to include in license generation.
       # Defaults to ["runtime", "runtimeClasspath"]
       def configurations
         @configurations ||= begin
-          if configurations = config.dig("gradle", "configurations")
-            Array(configurations)
-          else
-            DEFAULT_CONFIGURATIONS
-          end
+                              if configurations = config.dig("gradle", "configurations")
```
jonabc: revert this indenting as well?

jandersson-svt:
I agree on all you comments. Sorry for the formatting, and for forgetting the tests, I did this patch on a train ride and stressed to get it done before I reached home. I'll try to get the fixes done on friday, when we have our hack-on-whatever-you-want day. And of course, if anyone else should fell like supporting fixes before that, they are all welcome.

jandersson-svt:
A question about test fixture - how detailed do you want them - should I go for having all the variants we want to test in different fixture files. 1) no buildscripts or plugins block (like now) 2) only plugins block, no license dep added 3) only plugins block, license dep added   and so on. with every permutation/possible combination it would quite a few tests.

jonabc:
@jandersson-svt 

> Sorry for the formatting, and for forgetting the tests, I did this patch on a train ride and stressed to get it done before I reached home.

Not a problem at all!  Thank you for trying out the new source and putting together a PR to fix the issues you found 🙇 

> A question about test fixture(s) - how detailed do you want them ... I interpret the fixtures to be more of a high perspective integration tests though, and as a first step add the ones I think is most important. A few tests is better than no tests.

✨ this sounds great.  I'm not terribly concerned about covering all edge cases right now.  Getting coverage for common gradle files patterns would be perfect.

jandersson-svt:
Let's try this. Fixed all your suggestions as wished, I only did at simple addition to the build.gradle file for testing the pattern with a plugins block and a buildscript block. I'll add more later, but hope this basic is enough for this PR.

Btw, me and my colleague have some ideas and some suggestions, should I make an issue for each so that it can be discussed there? 

 (I also found one bug, but I think that is in one of the helper libraries, I'll have a look before making an issue or PR)

Issue #155, opened 2019-03-29
Include license information in standalone distribution


jclem:
Currently, the standalone distribution of licensed only contains the licensed binary itself. It would be helpful if the tarball included the license file, as well as license information for licensed's dependencies.

mlinksva:
Ideal thing would be a `NOTICE` file which contains dependencies and license info that licensed enumerates for itself. 😄 🐶 🥘

jonabc:
> Ideal thing would be a NOTICE file which contains dependencies and license info that licensed enumerates for itself.

👍 dogfooding!  What would you recommend the `NOTICE` file contents looking like in terms of data and formatting?

jonabc:
For now, I've updated the 2.0.1 release archives to contain ruby's license.txt, licensed's own LICENSE and README.md, and the *.dep.yml files from running licensed on itself.

```
➜  licensed git:(master) ✗ tar -tzf licensed-2.0.1-darwin-x64.tar.gz
./licensed
./meta/
./meta/LICENSE
./meta/README.md
./meta/.licenses/
./meta/ruby/
./meta/ruby/license.txt
./meta/.licenses/bundler/
./meta/.licenses/bundler/public_suffix.dep.yml
./meta/.licenses/bundler/dotenv.dep.yml
./meta/.licenses/bundler/addressable.dep.yml
./meta/.licenses/bundler/thor.dep.yml
./meta/.licenses/bundler/octokit.dep.yml
./meta/.licenses/bundler/licensee.dep.yml
./meta/.licenses/bundler/multipart-post.dep.yml
./meta/.licenses/bundler/faraday.dep.yml
./meta/.licenses/bundler/pathname-common_prefix.dep.yml
./meta/.licenses/bundler/rugged.dep.yml
./meta/.licenses/bundler/sawyer.dep.yml
./meta/.licenses/bundler/tomlrb.dep.yml
./meta/.licenses/bundler/bundler.dep.yml
```

I'll use this issue to track package scripting changes to make this automatic in the future.  Should previous releases be updated as well?

mlinksva:
Iff it's easy to update the previous releases, yes.

Regarding `NOTICE` formatting, I don't know of existing conventions off the top of my head, other than ad hoc concatenation of license notices sometimes with ad hoc metadata. So simple concatenation would be a reasonable first implementation.

jonabc:
[...]

PR #160, opened 2019-04-11
Include dependency license data in distributed packages

jonabc:
resolves https://github.com/github/licensed/issues/155

This updates the packaging scripts to include dependency data in built distributable packages.  Unfortunately the resulting script is somewhat fragile due to needing the environment _just right_ for both the docker and non-docker build environments.  I've attempted to comment the script for some of the pitfalls but it's hard to describe all the ways that things _just wont work_ using the prebuilt ruby docker image.  Sadly things work even less when using a custom docker image that installs ruby in a different manner.

In a short-ish description, the script works by
- creating a copy of the local repo in its current state without Gemfile.lock, existing bundled gems, existing dependency metadata and tests.
   - removing any installed state allows later bootstrapping and licensed execution to find the latest gems available, which is what is used in the executable.
- checking out the copied repo to the desired version if needed and building the executable
- restoring back to the current repo version, licensed is bootstrapped and run on itself
   - the environment needs to be _just right_ here, with at least a `>= 2.0` version of bundler available at `GEM_HOME`, no `Gemfile.lock`, and no `BUNDLER_VERSION` env var set
- lastly remaining metadata is pulled together from licensed's own `LICENSE` and `README.md` as well as ruby's `license.txt`

All of the dependency data is put under a `meta` folder in the package, with a resulting structure like
```
./licensed
./meta/
./meta/LICENSE
./meta/README.md
./meta/.licenses/
./meta/ruby/
./meta/ruby/license.txt
./meta/.licenses/bundler/
./meta/.licenses/bundler/public_suffix.dep.yml
./meta/.licenses/bundler/dotenv.dep.yml
./meta/.licenses/bundler/addressable.dep.yml
./meta/.licenses/bundler/thor.dep.yml
./meta/.licenses/bundler/octokit.dep.yml
./meta/.licenses/bundler/licensee.dep.yml
./meta/.licenses/bundler/multipart-post.dep.yml
./meta/.licenses/bundler/faraday.dep.yml
./meta/.licenses/bundler/pathname-common_prefix.dep.yml
./meta/.licenses/bundler/rugged.dep.yml
./meta/.licenses/bundler/sawyer.dep.yml
./meta/.licenses/bundler/tomlrb.dep.yml
./meta/.licenses/bundler/bundler.dep.yml
```

/cc @mlinksva for 👍 on the resulting structure and overall process.


jonabc:
> If I understand script correctly, rubyc and licensed are each running in clean environments (no gems installed yet) so can be expected to get the same gems/versions as long as there's nothing platform specific.

👍  that's my expectation and observation.  I think we're ok for platform specific gems with this approach as well.

PR #165, opened 2019-05-08
Python improvements

krzysztof-pawlik-gat:
- Added ability to handle packages that have `-` in their name, for example `ua-parser`
- Fixed handling of URLs in `requirements.txt` file (those are ignored for now)

### DIFF
--- a/lib/licensed/sources/pip.rb
+++ b/lib/licensed/sources/pip.rb
@@ -35,6 +35,7 @@ module Licensed
       def packages_from_requirements_txt
         File.read(@config.pwd.join("requirements.txt"))
             .lines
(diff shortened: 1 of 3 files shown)


jonabc:
:wave: @krzysztof-pawlik-gat this looks great 🙇 .

Would you mind adding a test and updating the [requirements.txt fixture](https://github.com/github/licensed/blob/afba288df344e001d43e94ad9217635a011b79c2/test/fixtures/pip/requirements.txt) to verify the behavior is fixed and prevent a regression in the future?

krzysztof-pawlik-gat:
Test added & build passing.

Issue #117, opened 2019-01-03
Pipenv support (Python)


dideler:
[Pipenv](https://pipenv.readthedocs.io/en/latest/) is a popular package and virtualenv manager for Python.

Like Ruby's Bundler which has `Gemfile` and `Gemfile.lock`, Pipenv has `Pipfile` and `Pipfile.lock`.

Note that Pipenv can produce a `requirements.txt` file, which I tried to use with licensed, but no luck.

```
$ pipenv lock --requirements > requirements.txt

$ head requirements.txt
-i https://pypi.org/simple
-e .
absl-py==0.6.1
astor==0.7.1
boto3==1.9.62
boto==2.49.0
botocore==1.12.62
bz2file==0.98
certifi==2018.11.29
chardet==3.0.4

$ licensed list
Displaying dependencies for datascience core package

$ cat .licensed.yaml
sources:
  pip: true
```

dideler:
Read [the licensed pip docs](https://github.com/github/licensed/blob/master/docs/sources/pip.md) and specified the virtualenv directory in `.licensed.yaml`.

```
sources:
  pip: true
python:
  virtual_env_dir: "/Users/dennis/.local/share/virtualenvs/core-06DYXQgl"
```

Trying again errors with the following trace

```
$ licensed list
Displaying dependencies for datascience core package
  pip dependencies:
Traceback (most recent call last):
	22: from /Users/dennis/.rbenv/versions/2.6.0/bin/licensed:23:in `<main>'
	21: from /Users/dennis/.rbenv/versions/2.6.0/bin/licensed:23:in `load'
	20: from /Users/dennis/.rbenv/versions/2.6.0/lib/ruby/gems/2.6.0/gems/licensed-1.5.2/exe/licensed:5:in `<top (required)>'
	19: from /Users/dennis/.rbenv/versions/2.6.0/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/base.rb:466:in `start'
	18: from /Users/dennis/.rbenv/versions/2.6.0/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor.rb:387:in `dispatch'
	17: from /Users/dennis/.rbenv/versions/2.6.0/lib/ruby/gems/2.6.0/gems/thor-0.20.3/lib/thor/invocation.rb:126:in `invoke_command'
[...]

PR #167, opened 2019-05-14
Pipenv source

krzysztof-pawlik-gat:
Adding a simple source for `pipenv` -- nice wrapper around `virtualenv` and good, old `pip`.

This merge request contains one non-pipenv related change: a fix for reading & storing notice files that are not UTF-8 encoded.

This will solve issue #117.

### DIFF
--- a/docs/sources/pipenv.md
+++ b/docs/sources/pipenv.md
@@ -1,0 +1,5 @@
+# Pipenv
+
+The pipenv source uses `pipenv` CLI command to enumerate dependencies and properties.
+
+Be sure to run `pipenv update` (or `pipenv sync`) before running `licensed` so all required packages are properly installed.
--- a/lib/licensed/dependency.rb
+++ b/lib/licensed/dependency.rb
@@ -83,11 +83,15 @@ module Licensed
          .grep(LEGAL_FILES_PATTERN)
          .select { |path| File.file?(path) }
          .sort # sorted by the path
-         .map { |path| { "sources" => normalize_source_path(path), "text" => File.read(path).rstrip } }
+         .map { |path| { "sources" => normalize_source_path(path), "text" => read_file_with_encoding_check(path) } }
          .select { |notice| notice["text"].length > 0 } # files with content only
     end
 
     private
+
+    def read_file_with_encoding_check(file_path)
+      File.read(file_path).encode("UTF-16", invalid: :replace, replace: "?").encode("UTF-8").rstrip
+    end
 
     # Returns the sources for a group of license file contents
     #
--- a/lib/licensed/sources/pipenv.rb
+++ b/lib/licensed/sources/pipenv.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+module Licensed
+  module Sources
+    class Pipenv < Source
+      def enabled?
+        Licensed::Shell.tool_available?("pipenv") && File.exist?(@config.pwd.join("Pipfile.lock"))
+      end
+
+      def enumerate_dependencies
+        pakages_from_pipfile_lock.map do |package_name|
+          package = package_info(package_name)
+          location = File.join(package["Location"], package["Name"].gsub("-", "_") +  "-" + package["Version"] + ".dist-info")
+          Dependency.new(
+            name: package["Name"],
+            version: package["Version"],
+            path: location,
+            metadata: {
+              "type"        => Pipenv.type,
+              "summary"     => package["Summary"],
+              "homepage"    => package["Home-page"]
+            }
+          )
+        end
+      end
+
+      private
+
+      def pakages_from_pipfile_lock
+        Licensed::Shell.execute("pipenv", "run", "pip", "list")
+            .lines
+            .drop(2)  # Header
+            .map { |line| line.strip.split.first.strip }
+      end
+
+      def package_info(package_name)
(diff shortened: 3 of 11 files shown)


test/sources/pipenv_test.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+    end
+
+    describe "dependencies" do
+      it "detects top-level dependencies" do
+        Dir.chdir fixtures do
+          dep = source.dependencies.detect { |d| d.name == "pylint" }
+          assert dep
+          assert_equal "pipenv", dep.record["type"]
+          assert dep.record["homepage"]
```
jonabc: can you also add some verifications to these tests that `dep.version` is the expected value?

lib/licensed/sources/pipenv.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module Licensed
+  module Sources
+    class Pipenv < Source
+      def enabled?
+        Licensed::Shell.tool_available?("pipenv") && File.exist?(@config.pwd.join("Pipfile.lock"))
```
jonabc: for clarification since I don't have much experience in this ecosystem - does this have potential overlap with the `pip` source?  Is there likely to be both a `Pipfile.lock` and `requirements.txt` in the same repo?

jonabc: Also, in the pip source the `pip` executable that's used is found based on configuration settings.  I'm assuming that's not needed here because `pipenv` knows how to find the right executable, is that correct?

docs/sources/pipenv.md
```diff
@@ -0,0 +1,5 @@
+# Pipenv
+
+The pipenv source uses `pipenv` CLI command to enumerate dependencies and properties.
+
+Be sure to run `pipenv update` (or `pipenv sync`) before running `licensed` so all required packages are properly installed.
```
jonabc: just to double check, there aren't any edge cases that users need to be aware of when using this source?  It all looks pretty straightforward but worth asking 😄

lib/licensed/dependency.rb
```diff
@@ -83,12 +83,16 @@ def notice_contents
...
          .select { |path| File.file?(path) }
          .sort # sorted by the path
-         .map { |path| { "sources" => normalize_source_path(path), "text" => File.read(path).rstrip } }
+         .map { |path| { "sources" => normalize_source_path(path), "text" => read_file_with_encoding_check(path) } }
          .select { |notice| notice["text"].length > 0 } # files with content only
     end
 
     private
 
+    def read_file_with_encoding_check(file_path)
```
jonabc: Is this something you've hit in the real world?  I'm curious what the scenario is that files contain invalid characters, and where `licensed` is raising errors.

lib/licensed/sources/pipenv.rb
```diff
@@ -0,0 +1,46 @@
...
+          )
+        end
+      end
+
+      private
+
+      def pakages_from_pipfile_lock
+        Licensed::Shell.execute("pipenv", "run", "pip", "list")
+            .lines
+            .drop(2)  # Header
```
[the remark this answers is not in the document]
krzysztof-pawlik-gat: The header comes from `pip` not from `pipenv`, hence this option does not apply here.

lib/licensed/dependency.rb
krzysztof-pawlik-gat: Sadly yes -- `AUTHORS` file for https://pypi.org/project/beautifulsoup4/ (version 4.7.1) is raising non-UTF-8 string exception.

docs/sources/pipenv.md
krzysztof-pawlik-gat: Yes, it's that simple. Where the simplicity comes from is what `pipenv` is. For `pip` one has to create the virtual environment & install the packages manually, then the `pip` source goes over what's in `requirements.txt` file. For `pipenv` it manages both the Python virtual environment & packages, hence having it setup everything (using `update` or `sync`) will yield a ready for use virtual environment with installed packages. After that we're just using `pipenv` to invoke the right `pip` to get list of all installed packages and get everything from it.

test/sources/pipenv_test.rb
krzysztof-pawlik-gat: Good point, added in next commit.

lib/licensed/sources/pipenv.rb
krzysztof-pawlik-gat: Yes, `pipenv` will run the correct `pip` (the one that's in the virtual environment managed by `pipenv`).

For the first comment (about both `Pipfile.lock` and `requirements.txt` coexisting) -- theoretically there is such possibility to have both, but it does not make much sense, `requirements.txt` is used in non-Pipenv situations (like no venv at all, global packages, manual `virtualenv`) while `Pipfile.lock` is exclusive to `pipenv`. If both are present user would probably like to select one or the other source for `licensed`, otherwise, he/she will get output from both (which does not do any harm in my opinion anyway).

krzysztof-pawlik-gat:
ghc/cabal test failed (due to timeout during `ghc` installation), I don't see a quick way of restarting the build on TravisCI (it may be my permissions thing). Can you restart the build? I'd like to avoid creating artificial commits just to trigger it.

PR #170, opened 2019-07-26
Reject packages that do not have path.

krzysztof-pawlik-gat:
This could happen for `package.json` files for which `npm` correctly
shows an error, but `yarn` works with them just fine.

As it is not a correct `npm` case it's not trivial to add a test for this case (as it will be rejected by `npm`), this will become a moot issue when proper `yarn` source is introduced. This happened in a case where there was a dependency conflict, one of the libraries deep down required specific version of `lodash` and at top level we require the latest one. The error was happening for the version deep down, `npm` produced following list in JSON:

```json
        "lodash": {
          "requiredBy": "4.6.1",
          "missing": true,
          "optional": false
        }
```

Note the lack of `version` and `path`, but it has `missing` set.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,19 @@ matrix:
       node_js: "8"
       before_script: ./script/source-setup/npm
       script: ./script/test npm
-      env: NAME="npm"
+      env: NAME="npm 8"
+
+    - language: node_js
+      node_js: "10"
+      before_script: ./script/source-setup/npm
+      script: ./script/test npm
+      env: NAME="npm 10"
+
+    - language: node_js
+      node_js: "12"
+      before_script: ./script/source-setup/npm
+      script: ./script/test npm
+      env: NAME="npm 12"
 
     # bower tests
     - language: node_js
--- a/lib/licensed/sources/npm.rb
+++ b/lib/licensed/sources/npm.rb
@@ -13,6 +13,7 @@ module Licensed
       end
 
       def enumerate_dependencies
+        @yarn_lock_present = File.exist?(@config.pwd.join("yarn.lock"))
         packages.map do |name, package|
           path = package["path"]
           Dependency.new(
@@ -47,9 +48,8 @@ module Licensed
       # Recursively parse dependency JSON data.  Returns a hash mapping the
       # package name to it's metadata
       def recursive_dependencies(dependencies, result = {})
-        yarn_lock_present = File.exist?(@config.pwd.join("yarn.lock"))
         dependencies.each do |name, dependency|
-          next if yarn_lock_present && dependency["missing"]
+          next if @yarn_lock_present && dependency["missing"]
           (result[name] ||= []) << dependency
           recursive_dependencies(dependency["dependencies"] || {}, result)
         end
--- a/lib/licensed/sources/npm.rb
+++ b/lib/licensed/sources/npm.rb
@@ -13,7 +13,7 @@ module Licensed
       end
 
       def enumerate_dependencies
-        packages.map do |name, package|
+        packages.reject { |_, pkg| pkg["path"].nil? }.map do |name, package|
           path = package["path"]
           Dependency.new(
             name: name,
--- a/lib/licensed/sources/npm.rb
+++ b/lib/licensed/sources/npm.rb
@@ -13,7 +13,7 @@ module Licensed
       end
 
       def enumerate_dependencies
-        packages.reject { |_, pkg| pkg["path"].nil? }.map do |name, package|
+        packages.map do |name, package|
           path = package["path"]
           Dependency.new(
             name: name,
@@ -48,6 +48,7 @@ module Licensed
       # package name to it's metadata
       def recursive_dependencies(dependencies, result = {})
         dependencies.each do |name, dependency|
+          next unless dependency["missing"].nil?
           (result[name] ||= []) << dependency
           recursive_dependencies(dependency["dependencies"] || {}, result)
         end
--- a/lib/licensed/sources/npm.rb
+++ b/lib/licensed/sources/npm.rb
@@ -47,8 +47,9 @@ module Licensed
       # Recursively parse dependency JSON data.  Returns a hash mapping the
       # package name to it's metadata
       def recursive_dependencies(dependencies, result = {})
+        yarn_lock_present = File.exist?(@config.pwd.join("yarn.lock"))
         dependencies.each do |name, dependency|
-          next unless dependency["missing"].nil?
+          next if yarn_lock_present && dependency["missing"]
           (result[name] ||= []) << dependency
           recursive_dependencies(dependency["dependencies"] || {}, result)
         end
--- a/script/source-setup/npm
+++ b/script/source-setup/npm
@@ -15,3 +15,16 @@ if [ "$1" == "-f" ]; then
 fi
 
 npm install
+
+# setup test fixtures with missing dependencies
+mkdir -p "${BASE_PATH}/test/fixtures/npm-missing"
+cd "${BASE_PATH}/test/fixtures/npm-missing"
+cp ../npm/package.json .
+
+if [ "${1}" == "-f" ]; then
+  find . -not -regex "\.*" -and -not -name "package\.json" -print0 | xargs -0 rm -rf
+fi
+
+npm install
+rm -rf node_modules/glob
+touch yarn.lock
--- a/script/source-setup/npm
+++ b/script/source-setup/npm
@@ -17,14 +17,7 @@ fi
 npm install
 
 # setup test fixtures with missing dependencies
-mkdir -p "${BASE_PATH}/test/fixtures/npm-missing"
-cd "${BASE_PATH}/test/fixtures/npm-missing"
-cp ../npm/package.json .
+mkdir -p missing-dependencies
+cp -R node_modules package* missing-dependencies/
 
-if [ "${1}" == "-f" ]; then
-  find . -not -regex "\.*" -and -not -name "package\.json" -print0 | xargs -0 rm -rf
-fi
-
-npm install
-rm -rf node_modules/glob
-touch yarn.lock
+rm -rf missing-dependencies/node_modules/glob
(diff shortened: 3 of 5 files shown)


jonabc:
:wave: @krzysztof-pawlik-gat thanks for opening this up!

I'm not terribly familiar with `yarn`'s requirements or what is considered an error case.  Is a missing path/dependency not considered an error for yarn?

I'm hesitant about this approach because it's either hiding an error that could be reported to the user or it's hiding a second version of a dependency used in the product.  For full dependency reporting both the top level version of `lodash` as well as the missing `4.6.1` should be listed in a bill of materials.

> this will become a moot issue when proper yarn source is introduced

I took a quick look at a yarn source a few months ago and couldn't find a way to easily enumerate all the information that licensed needs.  If you're familiar with the yarn CLI and how to pull package metadata and license information I would gladly accept a PR for this.

krzysztof-pawlik-gat:
This is not exactly `yarn`'s fault -- it does show a warning for this:

```
[1/4] 🔍  Resolving packages...
warning Resolution field "lodash@4.17.15" is incompatible with requested version "lodash@4.6.1"
warning Resolution field "lodash@4.17.15" is incompatible with requested version "lodash@4.17.14"
[2/4] 🚚  Fetching packages...
```

But it continues and installs the latest version, `npm` sees the package as `missing` (note that it's even marked as such). I've changed the code to skip `missing` packages instead of those without `path`, I hope this makes more sense and looks better.

jonabc:
@krzysztof-pawlik-gat sorry for the back and forth on this, I'm still concerned about a few things.

> warning Resolution field "lodash@4.17.15" is incompatible with requested version "lodash@4.6.1"
> warning Resolution field "lodash@4.17.15" is incompatible with requested version "lodash@4.17.14"

What is the nature of this warning?  Is it an error/warning that the yarn community at large ignores or do you only want to ignore it for your use case?  Are `lodash@4.6.1` and `lodash@4.17.4` ever installed and included in production/external deployments of the project?

My other concern is that the change is adding yarn-specific functionality that will end up affecting users that aren't using yarn.  Would it make sense to ignore missing packages only when a `yarnfile.lock` is present, as a differentiator between yarn and npm users?

krzysztof-pawlik-gat:
> @krzysztof-pawlik-gat sorry for the back and forth on this, I'm still concerned about a few things.

No problem at all, that's the purpose of reviews :)

> > warning Resolution field "lodash@4.17.15" is incompatible with requested version "lodash@4.6.1"
> > warning Resolution field "lodash@4.17.15" is incompatible with requested version "lodash@4.17.14"
> 
> What is the nature of this warning? Is it an error/warning that the yarn community at large ignores or do you only want to ignore it for your use case?

I've seen such warnings a few times, while it's good to not have them, it looks like it's not extremely important, as ...

> Are `lodash@4.6.1` and `lodash@4.17.4` ever installed and included in production/external deployments of the project?

No, those old (and vulnerable) versions are not installed, only the latest one (`4.17.15`) ends up in `node_modules`:

```console
$ find node_modules/ -name lodash
node_modules/lodash
```

From my testing it looks like the `missing: true` will be when packages are installed by `yarn`, but `npm` is used to list them as npm will happily install multiple versions of same package:

```console
$ find node_modules/ -name lodash
node_modules/graphql-toolkit/node_modules/lodash
node_modules/lodash
node_modules/gat/node_modules/lodash
```

> My other concern is that the change is adding yarn-specific functionality that will end up affecting users that aren't using yarn. Would it make sense to ignore missing packages only when a `yarnfile.lock` is present, as a differentiator between yarn and npm users?

Good point, I've added a check for `yarn.lock` before discarding the missing dependency.

lib/licensed/sources/npm.rb
```diff
@@ -47,7 +47,9 @@ def packages
       # Recursively parse dependency JSON data.  Returns a hash mapping the
       # package name to it's metadata
       def recursive_dependencies(dependencies, result = {})
+        yarn_lock_present = File.exist?(@config.pwd.join("yarn.lock"))
```
jonabc: Could you move this to a separate method and cache the result so that the file isn't looked at for every call to `recursive_dependencies`? :bow:

krzysztof-pawlik-gat:
Added tests for this case.

.travis.yml
```diff
@@ -47,7 +47,19 @@ matrix:
       node_js: "8"
       before_script: ./script/source-setup/npm
       script: ./script/test npm
-      env: NAME="npm"
+      env: NAME="npm 8"
```
jonabc: ❤️

script/source-setup/npm
```diff
@@ -15,3 +15,16 @@ if [ "$1" == "-f" ]; then
...
+mkdir -p "${BASE_PATH}/test/fixtures/npm-missing"
+cd "${BASE_PATH}/test/fixtures/npm-missing"
+cp ../npm/package.json .
+
+if [ "${1}" == "-f" ]; then
+  find . -not -regex "\.*" -and -not -name "package\.json" -print0 | xargs -0 rm -rf
+fi
+
+npm install
+rm -rf node_modules/glob
```
jonabc: It looks like these fixtures would only be used for a few tests.  Rather than introducing a new fixtures folder, I'd prefer that the test setup copies the npm files to a tmp folder.  Either the test setup or a test case itself could then remove packages to simulate the missing dependency

test/sources/npm_test.rb
```diff
@@ -80,5 +80,16 @@
         end
       end
     end
+
+    describe "missing dependencies" do
```
jonabc: This should also include a test that the non-yarn error scenario still occurs when `yarn.lock` is not present 🙇

script/source-setup/npm
```diff
@@ -15,3 +15,16 @@ if [ "$1" == "-f" ]; then
...
+touch yarn.lock
```
jonabc: This should be moved inside of the test case that verifies behavior when yarn is used.

.travis.yml
krzysztof-pawlik-gat: :)

script/source-setup/npm
krzysztof-pawlik-gat: Ok, got it -- I'm creating `missing-dependencies` directory in `npm` to avoid extra fixture, also just copying the required files from the level above.

script/source-setup/npm
krzysztof-pawlik-gat: Done.

test/sources/npm_test.rb
krzysztof-pawlik-gat: Added & generally improved those tests: they make sure that `yarn.lock` is present or missing (depending on method).

krzysztof-pawlik-gat:
@jonabc Travis shows a failure for Haskel/cabal, this was not touched at all in this change, can you trigger the build again? (I don't want to push empty commits).

krzysztof-pawlik-gat:
Comments addressed, tests passing, I think this is ready for another round of review :)

script/source-setup/npm
```diff
@@ -15,3 +15,9 @@ if [ "$1" == "-f" ]; then
 fi
 
 npm install
+
+# setup test fixtures with missing dependencies
+mkdir -p missing-dependencies
```
jonabc: 🤔 I'm not sure whether there should be an entirely new folder of fixtures for the two added test cases.  How do you feel about doing something similar to [these go tests](https://github.com/github/licensed/blob/e223ccab5f8299a278d4679c5c8c6e0982ce545e/test/sources/go_test.rb#L128-L151) or [this bundler tests](https://github.com/github/licensed/blob/e223ccab5f8299a278d4679c5c8c6e0982ce545e/test/sources/bundler_test.rb#L198-L209)?

test/sources/npm_test.rb
```diff
@@ -80,5 +80,27 @@
...
+          File.unlink("yarn.lock") if File.exist? "yarn.lock"
+
+          assert source.dependencies.detect { |dep| dep.name == "autoprefixer" }
+          assert source.dependencies.detect { |dep| dep.name == "glob" }
+        end
+      end
+
+      it "excludes missing dependencies when yarn.lock is present" do
+        Dir.chdir fixtures do
+          File.write("yarn.lock", "") unless File.exist? "yarn.lock"
```
jonabc: This yarn lockfile needs to be deleted as part of this test's cleanup.  It's 👍 that the other added test accounts for it, but none of the other tests in this file do and likely any other tests added in the future won't either.  It could cause unexpected behavior if the file sticks around past the end of the test 🙇

jonabc: Alternatively if you want to follow my other suggestion about copying the fixtures to a tmp folder and then manipulating the tmp folder to get the correct state for the test, this doesn't need to be cleaned up
```suggestion
          File.write("yarn.lock", "")
```

test/sources/npm_test.rb
```diff
@@ -80,5 +80,27 @@
         end
       end
     end
+
+    describe "missing dependencies (glob is missing package)" do
+      let(:fixtures) { File.expand_path("../../fixtures/npm/missing-dependencies", __FILE__) }
+
+      it "includes missing dependencies when yarn.lock is present" do
+        Dir.chdir fixtures do
+          File.unlink("yarn.lock") if File.exist? "yarn.lock"
```
jonabc: This test shouldn't need to clean up after the other test 👍 
```suggestion
```

script/source-setup/npm
krzysztof-pawlik-gat: Ok, followed bundler example (as it looks slightly cleaner to me).

test/sources/npm_test.rb
krzysztof-pawlik-gat: N/A after test rework.

krzysztof-pawlik-gat: N/A after test rework.

test/sources/npm_test.rb
krzysztof-pawlik-gat: N/A after test rework.

krzysztof-pawlik-gat:
Reworked tests to use a temporary directory (like `bundler` tests do`).

PR #184, opened 2019-09-15
Get cabal tests working more reliably

jonabc:
Some diagnostic output and fiddling with the cabal setup script and possibly tests to get it working reliably in [CI](https://github.com/github/licensed/runs/213761185).


jonabc:
Seems like maybe the minor script changes fixed it 🤷‍♂

PR #187, opened 2019-09-23
Add `env` command to output runtime environment configuration

jonabc:
This adds in a new command - `licensed env` which will output the configuration environment licensed uses during runtime.  The output can be formatted to json or yaml depending on a `--format` option, with yaml used by default.

Most of the data is given per-[app](https://github.com/github/licensed/blob/de0d4136ab2f1c0dc716454f3f9ef14f4fdf6271/docs/configuration.md#applications), with the
- name of the app
- full cache path
- full source path
- allowed licenses
- reviewed dependencies
- ignored dependencies
- which sources are being examined for that app
- the version strategy that's used when a package manager version isn't available (e.g. manifest source)
- any other configuration for the app that's either inherited from the root configuration or set directly on the app

examples 👇  are based off of this repos [`.licensed.yml`](https://github.com/github/licensed/blob/de0d4136ab2f1c0dc716454f3f9ef14f4fdf6271/.licensed.yml)

```yaml
---
root: "/Users/jonabc/github/licensed"
git_repo: true
apps:
- name: licensed
  cache_path: "/Users/jonabc/github/licensed/.licenses"
  source_path: "/Users/jonabc/github/licensed"
  allowed:
  - mit
  - apache-2.0
  reviewed:
    bundler:
    - pathname-common_prefix
  sources:
  - bower
  - bundler
  - cabal
  - composer
  - dep
  - git_submodule
  - go
  - manifest
  - npm
  - pip
  - pipenv
  - gradle
  ignored: {}
  version_strategy: git
```

```json
{
  "root": "/Users/jonabc/github/licensed",
  "git_repo": true,
  "apps": [
    {
      "name": "licensed",
      "cache_path": "/Users/jonabc/github/licensed/.licenses",
      "source_path": "/Users/jonabc/github/licensed",
      "allowed": [
        "mit",
        "apache-2.0"
      ],
      "reviewed": {
        "bundler": [
          "pathname-common_prefix"
        ]
      },
      "sources": [
        "bower",
        "bundler",
        "cabal",
        "composer",
        "dep",
        "git_submodule",
        "go",
        "manifest",
        "npm",
        "pip",
        "pipenv",
        "gradle"
      ],
      "ignored": {
      },
      "version_strategy": "git"
    }
  ]
}
```


jonabc:
This PR has a bit of churn associated with it.

In order to use the standard command/reporter structure, allow different reporters based on the `--format` flag and keep the logic internal to the command, I needed to make reporters created dynamically per command run rather than passed in as a constructor argument.

I also needed to let command `run*` overrides get access to the report object without necessarily having to re-implement the entire command.  The overrides can now pass a block to `super` to get access to the report object during evaluation.

```
super do |report|
  ...
end
```

Issue #186, opened 2019-09-19
Handle multiple licenses for the same project


jonabc:
I recently came across a case where a dependency was detected as both `mit` and `bsd-3-clause` which resulted in a classification of `other` and raised an error in `licensed status`.  However both of these license types were set as allowed in the configuration file, so there was no reason for this to be an error.

I'm wondering if it would be more appropriate then to store all detected license identifiers in  `.dep.yml` files and better support multi-licensed scenarios.

Some ideas off the top of my head

### 1. Store an array of licenses in a `license_keys`-like field

Storing multiple licenses in a new field and deprecating the `license` field would keep things backwards compatible and sane.  `licensed status` could be updated to check every `license_keys` entry if available, falling back to legacy behavior using `license` if unavailable.  

e.g.
```yaml
name: test
version: 1.0.0
license: other
license_keys: [mit, bsd-3-clause]
licenses:
- sources: LICENSE_MIT
   text: ...
- sources: LICENSE_BSD
   text: ...
```

The `licenses` key is already taken, but I'm sure there's something else that would make sense here to store all of the license keys found if `license_keys` is 👎 

The downside to this approach is that means potentially more in-depth manual examination and intervention if one of the multiple licenses is not detected properly, resulting in something like `[other, mit]`.

### 2. Include a `key` field on each `licenses` entry.

Adding a `key` field onto each entry in `licenses` more directly associates a license with each piece of text found.  `licensed status` would check all of these fields when available and default to the top level `license` field when unavailable.

As with the first option, the top level `license` field would be deprecated.

e.g.
```yaml
licenses:
- sources: LICENSE_MIT
   key: mit
   text: ...
- sources: LICENSE_BSD
   key: bsd-3-clause
   text: ...
```

The downside to this approach would be that the key data is a little harder to find for a human than a top level field, and the logic could be a little harder in deciding whether to check the deprecated `license` field vs the new `key` fields.

jonabc:
@mlinksva what do you think about ☝️ in general, and the various proposals.  Do you have any other ideas of how to store and check multiple licenses for each dependency?

mlinksva:
I think it's probably best to keep close to what licensee reports, which would be a mix of the two:

```yaml
name: test
version: 1.0.0
license: other
licenses:
- sources: LICENSE_MIT
   license: mit
   text: ...
- sources: LICENSE_BSD
   license: bsd-3-clause
   text: ...
```
[...]

PR #188, opened 2019-09-24
Check multiple licenses for allowed classifications 

jonabc:
Fixes https://github.com/github/licensed/issues/186

This PR updates `licensed status` to not raise an error in the case that 
1. The `license` field is set as `other`
2. There are multiple `licenses` entries
3. All `licenses` entries match a license classification that are configured `allowed`

This is meant to address cases where a project has multiple licenses that are each allowed and specified separately.  It won't address cases where a single `LICENSE` file references multiple licenses.

Under the hood this uses licensee to re-parse the cached content using both license file parsing and readme file parsing logic, and picking the first non-`other` license found between the two.  This should be fully backwards compatible as it doesn't require any additional metadata to work. 
 An assumption is made that if the content is parseable by both `LicenseFile` and `ReadmeFile`, that they won't result in different license classifications.


jonabc:
@mlinksva re: my last [comment](https://github.com/github/licensed/issues/186#issuecomment-534324174), I sidestepped the "which license classification do I edit" question by not adding any new fields, and evaluating the cached license text using licensee.

lib/licensed/commands/status.rb
```diff
@@ -35,20 +35,52 @@ def evaluate_dependency(app, source, dependency, report)
...
+        # review is not needed if the top level license is allowed
+        return false if app.allowed?(cached_record["license"])
+
+        # the remaining checks are meant to allow records marked as "other"
+        # that have multiple licenses, all of which are allowed
+
+        # review is needed for non-"other" licenses
+        return true unless cached_record["license"] == "other"
+
+        licenses = cached_record.licenses.map { |license| license_from_text(license.text) }
```
mlinksva: Ignorant question: can a license detected from a package manager file be in this list? Asking because it wouldn't be detected by `license_from_text` below.

mlinksva:
That sounds like a good strategy. I made one likely irrelevant comment.

lib/licensed/commands/status.rb
jonabc: 👍 it won't be detected because package manager text [isn't cached](https://github.com/github/licensed/blob/4abf047b99910f5a28a93f599381ab5bd7cdf8dc/lib/licensed/dependency.rb#L75) under the `licenses` field.

That in itself is somewhat confusing as it's a one-time / hidden check.  That's been around since before I started maintaining licensed, but I'll log an issue to re-examine if we should track that in the cached data and the best way to represent that.

jonabc: https://github.com/github/licensed/issues/189

PabloLION:
Hi, @jonabc. I know this is a old PR and released in 2.5.0, but I still want to ask if there is some example? 
I found that you wrote this https://github.com/github/licensed/issues/186#issuecomment-533236835 but I'm not sure it is the correct final version. Could you confirm it please?
Best wishes

jonabc:
@PabloLION :wave: the YAML syntax in the comment you linked isn't what ended up being used or implemented.  Individual files won't show their detected licenses.  This PR only changed the detection logic for when an "other" classification is determined as something the user needs to take some action on.

As an example, consider a project that references both BSD and MIT licenses, and both of those licenses are configured as `allowed` in the licensed configuration file.  Due to using two licenses, licensee will give the project a license classification of `other`.  Prior to this change that would have been called out as an error from `licensed status` and would have required the user to add the project to the `reviewed` licensed configuration.  With this change no error will be raised from `licensed status` and the user will not need to take any action.

This is referenced on the [status command documentation](https://github.com/github/licensed/blob/master/docs/commands/status.md), as the sub-bullet on case (4) for when the status command will raise an error.

PabloLION:
It seems I am asking in the wrong place. 
Would you mind telling me how to add two license so that they can be detected correctly? 

## tl;dr
These issues has the same problem and I'm referring them so people from SEO can see it.
https://github.com/gogs/gogs/issues/2173
https://github.com/github/choosealicense.com/issues/606

mlinksva:
@PabloLION this repo is **not** about the rule of how GitHub detects License of repos. https://github.com/licensee/licensee is the code for that. If there are multiple licenses present, GitHub won't detect any of them at the moment.

PabloLION:
Yes. I see it. Thank you. I agree to close the discussion here.

PR #195, opened 2019-09-30
Add Mix source

bruce:
Adds a new dependency source enumerator for `mix`, to enumerate Elixir package dependencies.

The source of information is a `mix.lock` file. In the event dependencies are installed (in `deps/`), a lockfile should be present, as the conventional `mix deps.get` command both downloads dependencies and generates the lockfile.

In order for dependencies to be enumerated:

- A `mix.lock` file must be available at the licensed app source path.

Technically speaking, since the `mix` command isn't called by licensed to extract metadata (rather, we parse the limited Elixir data structure that's written into `mix.lock`), an Elixir/Erlang/OTP runtime isn't needed when licensed is run.

cc @jclem

### DIFF
--- a/docs/sources/mix.md
+++ b/docs/sources/mix.md
@@ -19,3 +19,5 @@ Be sure to re-run this command whenever your `mix.exs` dependencies change to up
 ## Limitations
 
 Because `mix deps.get` does not generate `mix.lock` entries for `:path` dependencies (nor are they stored in `deps/`), license information is not extracted from them.
+
+If you need to extract license information from dependencies from another local directory (that is a Git repository), consider using a `:git` dependency with a file path in your `mix.exs`.
--- a/docs/sources/mix.md
+++ b/docs/sources/mix.md
@@ -1,0 +1,17 @@
+# Mix
+
+The mix source will detect `mix.lock` lockfiles, and parse their contents to enumerate dependencies.
+
+The `mix` CLI tool itself isn't used to extract dependency data, and no project compilation occurs during license extraction, but your dependencies should be present in your project `deps/` directory, as is conventional with Elixir projects.a
+
+## Installing Dependencies
+
+Your project should contain a `deps/ directory containing the dependency sources. You can download your dependencies (which are defined in `mix.exs`) using `mix`:
+
+```
+mix deps.get
+```
+
+This will create your `mix.lock` lockfile if needed.
+
+Be sure to re-run this command whenever your `mix.exs` dependencies change to update your lockfile and dependency sources.
--- a/docs/sources/mix.md
+++ b/docs/sources/mix.md
@@ -2,11 +2,11 @@
 
 The mix source will detect `mix.lock` lockfiles, and parse their contents to enumerate dependencies.
 
-The `mix` CLI tool itself isn't used to extract dependency data, and no project compilation occurs during license extraction, but your dependencies should be present in your project `deps/` directory, as is conventional with Elixir projects.a
+The `mix` CLI tool itself isn't used to extract dependency data, and no project compilation occurs during license extraction, but your dependencies should be present in your project `deps/` directory, as is conventional with Elixir projects.
 
 ## Installing Dependencies
 
-Your project should contain a `deps/ directory containing the dependency sources. You can download your dependencies (which are defined in `mix.exs`) using `mix`:
+Your project should contain a `deps/` directory containing the dependency sources. You can download your dependencies (which are defined in `mix.exs`) using `mix`:
 
 ```
 mix deps.get
--- a/docs/sources/mix.md
+++ b/docs/sources/mix.md
@@ -15,3 +15,7 @@ mix deps.get
 This will create your `mix.lock` lockfile if needed.
 
 Be sure to re-run this command whenever your `mix.exs` dependencies change to update your lockfile and dependency sources.
+
+## Limitations
+
+Because `mix deps.get` does not generate `mix.lock` entries for `:path` dependencies (nor are they stored in `deps/`), license information is not extracted from them.
--- a/lib/licensed/sources/mix.rb
+++ b/lib/licensed/sources/mix.rb
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-require "strscan"
 
 module Licensed
   module Sources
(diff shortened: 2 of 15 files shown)


tclem:
> cc @tclem

This looks cool, but maybe I'm not the @tclem you're looking for? Let me know if some specific feedback is required here!

bruce:
@tclem I appreciate the good humor (as just “bruce” I get pulled into a lot of random comments…), my apologies for the typo!

@jclem may find this of more specific interest 😉

lib/licensed/sources/mix.rb
```diff
@@ -0,0 +1,396 @@
...
+
+module Licensed
+  module Sources
+    class Mix < Source
+
+      LOCKFILE = "mix.lock"
+
+      # Returns whether a mix.lock is present
+      def enabled?
+        File.exists?(LOCKFILE)
```
jonabc: joining relative paths to `config.pwd` is preferable for clarity on the file location

```suggestion
        File.exists?(config.pwd.join(LOCKFILE))
```

lib/licensed/sources/mix.rb
```diff
@@ -0,0 +1,396 @@
...
+      # Returns a memoized Array of Dependency instances.
+      def parsed_dependencies
+        @parsed_dependencies ||= find_packages.map do |name, lock_info|
+          convert_package_to_dependency(name, lock_info)
+        end
+      end
+
+      # Returns the parsed mix.lock information as a Hash.
+      def find_packages
+        lockfile = File.read(LOCKFILE)
```
jonabc: nit: same here, mind using an absolute path?
```suggestion
        lockfile = File.read(config.pwd.join(LOCKFILE))
```

lib/licensed/sources/mix.rb
```diff
@@ -0,0 +1,396 @@
...
+      def git_dependency(name, lock_info)
+        # Example: {:git, "https://example.com/path/to/repo.git", "A-SHA-HERE", []},
+        path, errors = check_dep_path(name)
+        if lock_info.length == 4
+          Dependency.new(
+            name: name,
+            version: lock_info[2],
+            path: path,
+            metadata: {
+              "type" => "git",
```
jonabc: The `type` metadata should be an identifier for the package manager that this source enumerates.  There's a helper provided so that it doesn't need to be implemented per-class.  I should probably abstract this away entirely.
```suggestion
              "type" => self.class.type,
```

lib/licensed/sources/mix.rb
```diff
@@ -0,0 +1,396 @@
...
+            },
+            errors: errors
+          )
+        else
+          Dependency.new(
+            name: name,
+            path: path,
+            errors: errors << "unknown mix.lock format",
+            metadata: {
+              "type" => "git"
```
jonabc:
```suggestion
              "type" => self.class.type
```

lib/licensed/sources/mix.rb
```diff
@@ -0,0 +1,396 @@
...
+      def hex_dependency(name, lock_info)
+        # Example: {:hex, :pkgname, "1.2.3", "A-DIGEST-HERE", [:mix], [], "hexpm"}
+        path, errors = check_dep_path(name)
+        if lock_info.length == 7
+          Dependency.new(
+            name: name,
+            version: lock_info[2],
+            path: path,
+            metadata: {
+              "type" => "hex",
```
jonabc:
```suggestion
              "type" => self.class.type,
```

lib/licensed/sources/mix.rb
```diff
@@ -0,0 +1,396 @@
...
+            },
+            errors: errors
+          )
+        else
+          Dependency.new(
+            name: name,
+            path: path,
+            errors: errors << "unknown mix.lock format",
+            metadata: {
+              "type" => "hex"
```
jonabc:
```suggestion
              "type" => self.class.type
```

lib/licensed/sources/mix.rb
```diff
@@ -0,0 +1,396 @@
...
+      # Generate the absolute path to the named package.
+      #
+      # name - The name of the package dependency as a String.
+      #
+      # Returns a String.
+      def dep_path(name)
+        File.absolute_path(File.join(".", "deps", name))
+      end
+
+      class LockfileParser
```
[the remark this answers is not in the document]
bruce: I sure do love writing parsers—despite how painful they can be to others. 😆

Seriously, though, thanks for pointing that out. After a bit more review I think things are stable/structured enough for the approach you suggest. I took a stab at it, and even used some longform `//x` expanded regexes with comments to hopefully make things a little more understandable and maintainable. Let me know what you think.

bruce:
@jonabc Take another look when you have the time; I made the changes you requested (including the parser change). I'm capturing a couple of Mix-specific additional pieces of metadata about the dependencies (the `scm`, which is basically how Mix fetches the deps, and the `repo` which is the source); both pretty core pieces of info that are helpful when debugging.

docs/sources/mix.md
```diff
@@ -0,0 +1,21 @@
...
+mix deps.get
+```
+
+This will create your `mix.lock` lockfile if needed.
+
+Be sure to re-run this command whenever your `mix.exs` dependencies change to update your lockfile and dependency sources.
+
+## Limitations
+
+Because `mix deps.get` does not generate `mix.lock` entries for `:path` dependencies (nor are they stored in `deps/`), license information is not extracted from them.
```
jonabc: Hmmm are the `:path` dependencies enumerable from another source?  Not blocking to get this in, but if we're missing a class of dependencies entirely it's worth following up on

lib/licensed/sources/mix.rb
```diff
@@ -0,0 +1,177 @@
...
+        end
+
+        private
+
+        # Parse a line from the mix.lock file.
+        #
+        # line - A line of input as a String.
+        #
+        # Returns a Hash package entry, or raises a ParserError if unsuccessful.
+        def parse_line(line)
```
jonabc: Raising the custom `ParseError` will halt execution of the program and should generally be avoided.  

1. When `match` is not found - this looks like a `Source` level error, in that the `mix.lock` format is not understood and the source should halt execution.  

   You can handle `Source` errors by raising `Licensed::Sources::Source::Error`, which is handled by halting the execution of the current source and logging the error.

2. When `data` is not found - this looks like a `Dependency` level error, in that there is a name and scm match, but further data couldn't be found.

   `Dependency` errors are generally handled by setting the error message as `error` in `package_entry` which it looks like you already handle when creating the `Dependency` object.

docs/sources/mix.md
bruce: In practice `:path` dependencies are used in development only, usually for debugging. (Not to say that people don't build locally from path deps and release, but I haven't personally seen it.)

While it is possible to get the list of deps from `mix.exs` (the project description, used to generate `mix.lock`) via something like:

```shell
$ mix run --no-start PATH/TO/SCRIPT"
```

Where the contents of the script would be something like the following (but with more error-handling logic):

```elixir
Mix.Dep.load_on_environment([]) # maybe specifically :prod
|> Enum.map(fn
  %{scm: Hex.SCM, status: {:ok, _}} = dep ->
    # Extract data
  %{scm: Mix.SCM.Path, status: {:ok, _}} = dep ->
    # Extract data
  %{scm: Mix.SCM.Git, status: {:ok, _}} = dep ->
    # Extract data
  dep ->
    # Generate error information
end)
|> format_for_parsing()
```

I'd suggest waiting until someone actively asks for `:path` support, as this introduces a lot of complexity/brittleness if Mix's internals change over time.

bruce: One thing worth mentioning: if people use a local `:git` dependency vs `:path`, we get the definition in the `mix.lock`, e.g., changing the following in their `mix.exs`.

```
  {:absinthe, path: "/Users/bruce/code/absinthe-graphql/absinthe"}
``

to:

```
  {:absinthe, git: "/Users/bruce/code/absinthe-graphql/absinthe"}
```

Maybe I can suggest that as a possible work-around in the `mix.md` documentation?

lib/licensed/sources/mix.rb
bruce: 1. This makes sense, will, do.
2. I think the reason I didn't do this is because `version` and `path` was listed as required attributes (somewhere in the docs?). I assume I can just set `version` to `nil`, however?

bruce:
@jonabc I think I took care of the `.yml`, error handling concerns, and I added some additional testing as well. Let me know what you think about the comment on `:path`. (I think it's safe to document at the moment; I would be fairly surprised if someone even asks, considering how those deps are generally used.)

PR #200, opened 2019-10-05
Update commands.md

sergey-alekseev:
[ci skip]

PR #201, opened 2019-10-06
Update README.md

sergey-alekseev:
[ci skip]

Another iteration to https://github.com/github/licensed/pull/200.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,10 @@ For example, on macOS with Homebrew: `brew install cmake pkg-config` and on Ubun
 
 See the [commands documentation](./docs/commands.md) for additional documentation, or run `licensed -h` to see all of the current available commands.
 
+### Automation
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,6 +63,10 @@ For example, on macOS with Homebrew: `brew install cmake pkg-config` and on Ubun
 
 See the [commands documentation](./docs/commands.md) for additional documentation, or run `licensed -h` to see all of the current available commands.
 
+### Automation
+
+Do `bundle plugin install bundler-licensed` to automatically run `licensed cache` after running `bundle install` or `bundle update` commands.
```
jonabc: What do you think about this?  I'd rather link your repo with generic information than include specific details here.  If the plugin changes that reduces the need to update this README, is transparent to users where the plugin source is, and is more clear that the plugin is not "owned" by this repo

```suggestion
The [bundler-licensed plugin](https://github.com/sergey-alekseev/bundler-licensed) runs `licensed cache` automatically when using `bundler`.  See the linked repo for usage and details.
```

sergey-alekseev: Makes sense.

PR #204, opened 2019-10-22
Run pipenv in parallel for discovered packages to improve performance.

krzysztof-pawlik-gat:
Running `pipenv` in parallel significantly improved the performance of `pipenv` source, on my 8 core (4 cores + ht) machine running `./script/test pipenv` takes:

2.5.0:
 - 25.979721s
 - 25.930273s
 - 26.492875s

2.5.0 + parallel (threads):
 - 8.658091s
 - 8.751204s
 - 8.751976s

### DIFF
--- a/lib/licensed/sources/pipenv.rb
+++ b/lib/licensed/sources/pipenv.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "parallel"
+
 module Licensed
   module Sources
     class Pipenv < Source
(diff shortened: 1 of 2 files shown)


krzysztof-pawlik-gat:
> sparkles this is great, thank you! So great, in fact, that I'm going to extend it as a pattern to the other sources and possibly the parent `Source` class.

Sounds good, I think that `pip` source could use this too (from a quick glance), `npm` probably not as it invokes the binary once anyway. I didn't look at other sources, to be honest.

> I'm going pushing a new release and will merge this PR after. Due to CI not running on forks, I may ping you in an issue if this ends up with CI problems after merging to master.

I was surprised by that too! If there are any issues with this change I'll be happy to fix those.

PR #208, opened 2019-10-26
Add manifest tests to CI

jonabc:
Adds missing manifest source tests to the test workflow.  Must have missed this one when converting from travis to actions


jonabc:
Going to ignore the failing composer tests, and handle them in a separate PR.  Manifest tests are passing 👍

Issue #193, opened 2019-09-27
We should be able to ignore projects with wildcards


carlosmn:
I'm dealing with a Go project with `dep` and one of the internal projects I want to ignore contains several packages within it. I have to list every individual package we import instead of saying "github.com/github/someproject/*` can be ignored, I have to list each of the dozen packages in which it's split.

It feels like a lot of busywork not to be able to say "this project as a whole should be ignored". I get this might be trickier with other dependency mechanisms, but for those that are path-based, it would really save some set up time (and would more closely match what we'd do if there were packaging).

PR #214, opened 2019-10-29
Ignore list - support wildcard

gogiel:
Fixes https://github.com/github/licensed/issues/193

Refactoring note: I think `AppConfiguration` has too many responsibilities. It should only handle raw configuration and methods like `ignored?` should be moved. They are using the configuration, but not the part of the configuration. I didn't want to do that in this PR to make it more consistent. I'd be happy to do that in a separate PR.


gogiel:
Thank you for the instant response!

> This is quite a bit of additional code that could easily be solved by changing `Array#include?` to `Array#any?` + `String#match?`. Why all the extra abstraction and indirection?

I follow SOLID and made it future proof. Smaller classes make code easier to understand and what's more important to me to test in isolation. I'm able to test small components in full isolation.

I also wanted to make it future proof. As for now licensed supports only strings, but maybe in the future there's a case for more complex configuration.
E.g. in a case where one library has different license depending on version, etc.

I'm fine with removing most of the abstractions though, as that's not directly needed here.

> > Refactoring note: I think AppConfiguration has too many responsibilities. It should only handle raw configuration and methods like ignored? should be moved.
> 
> I strongly disagree with this. The `AppConfiguration` class is not just for storing raw data as is shown by its implementation. Why do you say otherwise? Can you give clear examples of how the refactoring you're suggesting make the product easier to write, easier to read, easier to maintain, or easier to use?

I'd say that it slightly starts to violate single responsibility principle with this new feature.
`AppConfiguration` main responsibility is handling raw YAML configuration with defaults, which is one responsibility.

The other responsibility that's more visible now is an action that uses this raw data.
`#ignored?` method was just a wrapper around `include?` but now it got more complex.

A good example of this paradigm is Rails configuration.

You configure your app with simple, e.g. `config.param = :something`. The configuration object is just dumb store with getters and setters and handling some default. It has no behaviour.

Behaviour is encapsulated in other modules which use the configuration object.

I was thinking about diving `AppConfiguration` into three independent parts:
* YAML/JSON handling on the *file* level - searching for default config files, parsing configs
* data store
* logic based on config, namely `#enabled?`, `#reviewed?`, `#ignored?`, `#allowed?` methods

jonabc:
> I also wanted to make it future proof. As for now licensed supports only strings, but maybe in the future there's a case for more complex configuration.
> E.g. in a case where one library has different license depending on version, etc.

As a general idea this makes sense 💯, but it sounds like the future needs aren't known here.  In the example that you're suggesting, it feels like an antipattern to configure licensed about a version of a dependency that's not being used.  Or if it is used, then the source should be giving them distinct names so that you can do something like

```yaml
# in this example two versions of dependency are referenced by npm - 4.0 and 5.0, where 4.0 uses a license not set as allowed, but 5.0 does
reviewed:
  npm:
  - dependency-4.0
```

jonabc:
Sorry, clicking all the wrong buttons 😆

gogiel:
Simplified it a bit.
I can join `MatchersList` and `WildcardMatcher` into one file if you think that's worth it.

jonabc:
@gogiel I'm going to close this for now - I've put together an [alternative PR](https://github.com/github/licensed/pull/225) that I'm going to move forward with.

As I said before I'm always happy to chat if you have thoughts on how to improve the tool or it's usages.  Thanks for opening the PR and contributing to the tool 🙇

PR #220, opened 2019-11-06
Make installation block more usable

Miroshni4enko:
When you install Licensed, you don't read the dependencies block because it is at the bottom of the installation paragraph. It caused a problem during installation. Let's make installation more user-friendly :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ See the [migration documentation](./docs/migrating_to_newer_versions.md) for mor
 
 ### Dependencies
 
-Install `cmake` and `pkg-config` dut to Licensed uses the `libgit2` bindings for Ruby provided by `rugged` which use these dependencies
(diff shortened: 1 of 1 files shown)


jonabc:
👍 makes sense.  Thanks for helping make this easier to use.

README.md
```diff
@@ -21,6 +21,18 @@ See the [migration documentation](./docs/migrating_to_newer_versions.md) for mor
 
 ## Installation
 
+### Dependencies
+
+Install `cmake` and `pkg-config` dut to Licensed uses the `libgit2` bindings for Ruby provided by `rugged` which use these dependencies
```
jonabc: Ah sorry I just noticed this. Can you use the pre-existing text, or maybe the suggested text below?

```suggestion
Licensed uses the `libgit2` bindings for Ruby provided by `rugged`. `rugged` requires `cmake` and `pkg-config` which you may need to install before you can install Licensed.
```

Miroshni4enko: Thanks!)

PR #221, opened 2019-11-07
Fix evaluate dependency bug

westoque:
This fixes a bug when adding dependencies that have empty licenses but have their allowed license in the `.licensed.yml`. This happens when we included the `aws-sdk-s3` gem:

our `.licensed.yml`:
```
sources:
  bundler: true
  npm: false
  bower: false
  cabal: false

allowed:
  - mit
  - apache-2.0

reviewed:
  bundler:
    - aws-eventstream
```

and our `.licenses/bundler/aws-eventstream.dep.yml` file:

```
---
name: aws-eventstream
version: 1.0.2
type: bundler
summary: AWS Event Stream Library
homepage: http://github.com/aws/aws-sdk-ruby
license: apache-2.0
licenses: []
notices: []
```

everytime we run `bundle exec licensed status`, we get:

```
* agencymvp.com.bundler.aws-eventstream
  filename: /Users/westoque/GitHub/westoque/rails_app/.licenses/bundler/aws-eventstream.dep.yml
    - missing license text
```

There's really no need to check if licenses are empty because we already do the empty check in `#license_needs_review?`.

### DIFF
--- a/lib/licensed/commands/status.rb
+++ b/lib/licensed/commands/status.rb
@@ -34,7 +34,6 @@ module Licensed
           report.errors << "cached dependency record not found"
         else
           report.errors << "cached dependency record out of date" if cached_record["version"] != dependency.version
-          report.errors << "missing license text" if cached_record.licenses.empty?
           report.errors << "license needs review: #{cached_record["license"]}" if license_needs_review?(app, cached_record)
         end
 


jonabc:
@westoque :wave: thanks for opening this PR!  This check is intentional in the product, at least at this point in time.  The problem here, particularly with _all_ of the AWS gems, is that they don't include license text in the distributed gem.  The right fix is to get the AWS gem maintainers to package a license file with their distributed gems.

Many licenses (apache-2.0 included) have clauses that specify the text of the license must be included for full compliance when using the dependency.

You can remediate these failures by manually adding the license text from the gem's source at the version you're using.  For the aws gems that would be [this version of LICENSE.txt](https://github.com/aws/aws-sdk-ruby/blob/138f9caf012a139fe7458e70f498e1311c6d80b0/LICENSE.txt).

We do our best to make this as infrequent of a pain point as possible, but the license text is needed for full compliance.  I'm going to close this but please let me know if you have any further questions 🙇

westoque:
@jonabc thanks for that wonderful explanation 👍

PR #222, opened 2019-11-09
Delete unneeded error raising

Miroshni4enko:
Looks like licensed doesn't need to raise an error without default config file. Even if you add an empty default file everything works and error during installation only make users "feel frustration":)


Miroshni4enko:
Thank you for so quick response)
I got your point and it sounds reasonable. Also, I thought that `licensed status` command has some default settings but it not and I think it's right. Users should deeply understand which license is allowed.

Miroshni4enko:
So, sorry for spend time and we can close this PR.

PR #225, opened 2019-11-10
Allow ignoring projects with wildcards

jonabc:
closes https://github.com/github/licensed/issues/193

This adds pattern matching when ignoring dependencies based on glob patterns and `File.fnmatch?`.  Comparisons use `File::FNM_PATHNAME` for strict `**` vs `*` evaluation, and `FNM_CASEFOLD` for case insensitivity.  For full details, please see the ruby documentation on `File.fnmatch?`

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -95,6 +95,12 @@ ignored:
   bower:
     - some-internal-package
 
+  go:
+    # ignore all go packages from import paths starting with github.com/internal-package
+    # see the `File.fnmatch?` documentation for details on how patterns are matched.
+    # comparisons use the FNM_CASEFOLD and FNM_PATHNAME flags
+    - github.com/internal-package/**/*
+
 # These dependencies have been reviewed.
 # They need to be cached and checked, but do not have a license found that matches the allowed configured licenses.
 reviewed:
--- a/lib/licensed/configuration.rb
+++ b/lib/licensed/configuration.rb
@@ -75,7 +75,9 @@ module Licensed
 
     # Is the given dependency ignored?
     def ignored?(dependency)
-      Array(self["ignored"][dependency["type"]]).include?(dependency["name"])
+      Array(self["ignored"][dependency["type"]]).any? do |pattern|
+        File.fnmatch?(pattern, dependency["name"], File::FNM_PATHNAME | File::FNM_CASEFOLD)
+      end
     end
 
     # Is the license of the dependency allowed?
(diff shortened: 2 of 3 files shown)


saviogl:
@jonabc Have you considered applying the same wildcard pattern to **allowed** dependencies too? I've got a node project here pulling in some (32) [HAPI](https://hapi.dev/) framework dependencies all of which have a `LICENSE.md` that for some reason is not matching against BSD-3-Clause, though it looks like it valid, and therefore I have to explicitly allow all of them in the `.licensed.yaml` file

jonabc:
@mlinksva :wave: thoughts on ☝️ ?  What is the level of concern for false-positive marking something as ok/reviewed?

@saviogl FWIW there are subtle differences between that text and the stock bsd-3 license which I think is throwing it off.  The general recommendation here is to manually update `license: other` to `license: bsd-3-clause`, which is admittedly an up front burden.  It should be a one-time cost though for as long as the license text remains static - the license classification will remain intact through version upgrades so long as the content doesn't change.

mlinksva:
I probably wouldn't recommend using a pattern in reviewed or ignored. If you care enough to use licensed, it doesn't strike me as a great tradeoff -- unless perhaps the invariant part is a namespace controlled by a project you have reason to expect consistency from -- maybe that is the case these days for `@hapi` -- but that seems like an edge case.

re subtle differences between the text and stock bsd-3-cluase -- "OFTWARE", wow 😆 but yes, even without that typo, the variation in the 3rd bullet would throw licensee off.

jonabc:
> unless perhaps the invariant part is a namespace controlled by a project you have reason to expect consistency from

👍  this is why ignore pattern matching exists.  It was built with go in mind, where it's common for `github.com/<owner>/<repo>/*` packages to all share a single license that's found at `github.com/<owner>/<repo>`

saviogl:
@jonabc @mlinksva I see where you are coming from and it's a valid indeed - I thougth to raise it up here as projects under an umbrella organization like `@hapi/*` could likely utilize code generators and standards that yields the same `License` scheme thus potentially beeing an argument to "accept" the wildcard pattern. With that said, there is a risk to it indeed and it might be better to force consumers to review all of its dependencies.

As a side note is there an effort to standardize licensing information accross open source to avoid these slightly mismacth?

mlinksva:
> 👍 this is why ignore pattern matching exists

Makes sense, ~struck~ my fragment about it being an edge case. Seems like a useful feature to be used with caution, including for reviewed.

Re standardize licensing effort, kind of (SPDX) but it doesn't usually play out in the form of existing projects modifying their license texts to be closer to canonical/most widely used text.

PR #233, opened 2019-12-30
Addressing some feedback from new yarn source

jonabc:
:wave: @sergey-alekseev @krzysztof-pawlik-gat thanks for taking a look at the [PR for the new yarn source](https://github.com/github/licensed/pull/232)!

I'd appreciate if you could also take a look at this PR and let me know if I'm addressing all of your comments and feedback 🙇 

1. protecting against extra json output from `yarn list` and `yarn info` by searching for the `tree` and `inspect` types, respectively.
2. moving `JSON.parse` outside the parallel handling from `Parallel.map`
   - I think general lack of thread-safety might have been the cause of @sergey-alekseev's [seen error](https://github.com/github/licensed/pull/232#issuecomment-569618866)
   - the reason to parallelize the calls is that `yarn info` is slow.  that call is still parallelized 👍 
3. nit cleanup on the yarn fixture

### DIFF
--- a/lib/licensed/sources/yarn.rb
+++ b/lib/licensed/sources/yarn.rb
@@ -39,7 +39,7 @@ module Licensed
             # if there is more than one package for a name, reference each by
             # "<name>-<version>"
             results.each do |package|
-              all_dependencies[package["id"].sub("@", "-")] = package
+              all_dependencies["#{name}-#{package["version"]}"] = package
             end
           end
         end
@@ -57,7 +57,7 @@ module Licensed
       def recursive_dependencies(path, dependencies, result = {})
         dependencies.each do |dependency|
           next if dependency["shadow"]
-          name, version = dependency["name"].split("@")
+          name, _, version = dependency["name"].rpartition("@")
 
           dependency_path = path.join("node_modules", name)
           (result[name] ||= []) << {
--- a/lib/licensed/sources/yarn.rb
+++ b/lib/licensed/sources/yarn.rb
@@ -44,7 +44,12 @@ module Licensed
           end
         end
 
-        Parallel.map(all_dependencies) { |name, dep| [name, package_info(dep)] }.to_h
+        # yarn info is a slow operation - run it in parallel for all dependencies.
+        # afterwards parse and merge the returned data serially, JSON.parse
+        # might not be thread safe
+        Parallel.map(all_dependencies) { |name, dep| [name, dep, yarn_info_command(dep["id"])] }
+                .map { |(name, dep, info)| [name, merge_yarn_info(dep, info)] }
+                .to_h
       end
 
       # Recursively parse dependency JSON data.  Returns a hash mapping the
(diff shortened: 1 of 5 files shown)


lib/licensed/sources/yarn.rb
```diff
@@ -26,22 +26,30 @@ def enumerate_dependencies
...
+        recursive_dependencies(config.pwd, yarn_package_tree).each do |name, results|
           results.uniq! { |package| package["version"] }
           if results.size == 1
+            # if there is only one package for a name, reference it by name
             all_dependencies[name] = results[0]
           else
+            # if there is more than one package for a name, reference each by
+            # "<name>-<version>"
             results.each do |package|
               all_dependencies[package["id"].sub("@", "-")] = package
```
krzysztof-pawlik-gat: I'm not entirely sure about this, for example, consider package named `@apollo/react-common`, with the version it becomes `@apollo/react-common@3.1.3`, so this line would cause the first `@` to be replaced too. Maybe something like `.sub(/(.)@(.)/, '\1-\2')` would work?

Especially that `npm` source does not strip the leading `@`. Check this branch: https://github.com/krzysztof-pawlik-gat/licensed/tree/ensure-at-in-packages (or compare view: https://github.com/github/licensed/compare/master...krzysztof-pawlik-gat:ensure-at-in-packages?expand=1).

krzysztof-pawlik-gat:
Thank you @jonabc, this looks good, I found one more issue but it's in review comments.

lib/licensed/sources/yarn.rb
jonabc: 🙏  great catch, I forgot about scoped packages.  updated in https://github.com/github/licensed/pull/233/commits/9005eda88cbfc0b6c9460ea508fe2af2436ed3db

sergey-alekseev:
Seems like [the error which I previously reported](https://github.com/github/licensed/pull/232#issuecomment-569618866) is gone 👍 

I tested the yarn source locally again and it seems to be working well. In comparison with the bundler source it takes quite a bit of time. See below.

When I run `licensed status` it immediately progresses with bundler dependencies and then stuck for some time without any progress with yarn dependencies. Finally it completes and prints everything at once. In my case:  
```
# with yarn source enabled
$ time licensed status
...
1584 dependencies checked, 198 errors found.

real	3m50.166s
user	9m12.936s
sys	1m30.953s

# with yarn source disabled
$ time licensed status
...
190 dependencies checked, 0 errors found.

real	0m0.972s
user	0m0.709s
sys	0m0.252s
```

`licensed cache` takes a lot of time too even when dependencies are already cached. In my case:  
```
# with yarn source enabled
$ time licensed cache
...
  * 1394 yarn dependencies

real	3m41.637s
user	9m30.477s
sys	1m33.880s

# with yarn source disabled
$ time licensed status
...
  * 190 bundler dependencies

real	0m1.304s
user	0m0.897s
sys	0m0.370s
```

---
Thanks for your work @jonabc! Happy New Year 😃

jonabc:
Yea the yarn source is really slow compared to bundler due to the extra shell execution of `yarn info` on each yarn dependency.  In your example that looks like around 1300 total calls to `yarn info`.  That's compared to npm which gets all of its info from a single `npm list`, or bundler which primarily uses the bundler API to get data, and so are much faster.

The extra call is used to grab the description and homepage.  The description isn't as important, but giving proper attribution for the homepage might be?  Not sure TBH

sergey-alekseev:
Thanks for the info @jonabc. I haven't figged into details and code, but my first thought is "can we avoid calling `yarn info` for `licensed status` and for already cached dependencies during `licensed cache`?"

jonabc:
@sergey-alekseev good idea!  That would require deeper changes than would make sense to include in this PR though 😞. 

Can you open an issue?  I have a few ideas on how I would structure the change that I can add to the new issue, but if you wanted to give that change a shot and open a PR I would gladly review it 😄

sergey-alekseev:
@jonabc I opened an issue regarding skipping `yarn info` as you requested – https://github.com/github/licensed/issues/235.

wadsaad:
Handle document

PR #234, opened 2020-01-01
add isc license to allowed by default

sergey-alekseev:
During testing the yarn source I found out a number of packages have the ISC license (98 out of 1394 in my case). E.g. https://github.com/isaacs/chownr/blob/v1.1.3/LICENSE.

[ci skip]

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -85,6 +85,7 @@ allowed:
   - bsd-2-clause
   - bsd-3-clause
   - cc0-1.0
+  - isc
 
 # These dependencies are explicitly ignored.
 # They shouldn't be cached at all, and will not have cached metadata written to the repo.


jonabc:
👍 to docs changes, but I think the intention of the example is being misunderstood.  There are no "allowed by default" licenses; users must always provide that list in their configuration files.

Perhaps the comment there, `Dependencies with these licenses are allowed by default.` should be changed?  Do you think that removing `by default` would be more clear that the comment is only meant as a description of what the items in that list are for, or are larger changes needed?

sergey-alekseev:
>users must always provide that list in their configuration files.   

Actually I think I never thought otherwise. However I see now after your comment this may be misleading for someone to think they may not configure `allowed:`. Anyway after the first `licensed status` failure they will realize they need to configure `allowed:`. So up to you to change or not 😉

jonabc:
Out of curiosity then, what was your purpose for adding the isc source to the example?  👍 to merge regardless, I think I misunderstood the intention and I'd love to understand better!

sergey-alekseev:
I just wanted to help others. Initially during setting up I copy-pasted those licenses from the doc to my config, I suppose others may do the same. That licenses list seems like a good default for most projects. Now when the yarn source is supported other users are likely to encounter the same failure as I had with ISC, read about this license and eventually add to their configuration list realizing it's similar to MIT. Hope this clarifies.

PR #237, opened 2020-01-03
Better descriptions for configuration settings in docs

jonabc:
From a number of discussions and opened issues, it's become obvious that some of the configuration settings purpose wasn't clear.

I've updated the descriptions to be a little more clear on both the purpose of the settings as well as how they affect licensed at runtime.


jonabc:
Going to merge this to get it into the release, 👍 to continue iterating on the docs and examples if there are things that could be improved or changed.

Issue #239, opened 2020-02-14
Enhancement request: allow multiple `apps` to be configured using wildcards


mhagger:
I'm working with a Go project that builds multiple executables. In my case, the main files are arranged like

* `cmd/foo/main.go`
* `cmd/bar/main.go`
* `cmd/baz/main.go`
* …

, which is common in the Go world. Currently it seems that I have to list these `apps` individually, like

```
apps:
  - source_path: cmd/foo
  - source_path: cmd/bar
  - source_path: cmd/baz
  - …
```

Aside from being wordy, this seems error-prone. It's quite plausible that somebody might add another compilation target but forget to adjust the `.licensed.yml` file.

It would be nice if apps could be configured using wildcards, like

```
apps:
  - source_path: cmd/*
```

/cc @jonabc

jonabc:
👍 I've started working on this but hit a snag in that I needed to change some of the underlying architecture a little.  Opened https://github.com/github/licensed/pull/240, after which this change should be relatively straightforward as a glob expansion of `source_path` values

PR #240, opened 2020-02-17
Remove Configuration < AppConfiguration inheritance

jonabc:
This PR makes little functional change, but impacts the majority of tests 😢 

Before the `apps` configuration option was available as a way to reference multiple targets in a repo, the `Configuration` class satisfied the functional role that the `AppConfiguration` class currently does.  When `AppConfiguration` was introduced, the `Configuration` class was made to inherit from `AppConfiguration` to keep the change minimally impactful to a large part of the codebase that used `AppConfiguration` methods but still only expected a single, top-level `Configuration`.

With the [request  to allow wildcard source paths](https://github.com/github/licensed/issues/239), it will no longer be possible to reference a `Configuration` as an `AppConfiguration`.  Where the `source_path` used to be treated as a string path that either existed or didn't, a wildcard `source_path` now needs to be evaluated and expanded into multiple `AppConfiguration`s.

It will be difficult to expand `source_path` at runtime due to cascading dependencies on the `source_path` value
- `AppConfiguration["name"]` defaults to being the right-most directory name of `source_path`
- `AppConfiguration#cache_path` relies on `AppConfiguration["name"]` when a `cache_path` value hasn't been explicitly set for the app

The logical way to expand a wildcard `source_path` is to create `AppConfiguration`s from each actualized value.  To keep things simple then, a root `source_path` would always evaluate into `AppConfiguration`s .  At this point a root `Configuration` that inherits from `AppConfiguration` becomes an architecture decision that only gets used in test code, which distances the test scenarios from production scenarios.

Following all of ☝️ reasoning, this PR drops the inheritance and updates a massive amount of test code to no longer reference a root level configuration

The command tests underwent a pretty major refactor.  it probably would have been alright to make minimal changes, but that would have been putting these changes off until later.  Since the majority of this change is test updates anyway it seemed like a good time to get it done.
   - added `sources` configuration to each `tests/fixtures/command/*.yml` file
   - when not loaded from a file, the configuration specifies to use the `TestSource` source type and allows for `source_config` options to be set as needed for individual tests.
   - with the above, it was no longer necessary to reference a specific `app` or `source` variable, and the tests were updated to use `config.apps.each` to test behavior for different app configurations.
   - I also needed to make some small modifications to the test command and test reporter to maintain assertion parity.

The remainder of the tests were fairly easy to convert.  Replacing `Licensed::Configuration` with `Licensed::AppConfiguration` fixed the majority of use cases, and fixing the handful of remaining issues that came up wasn't too complicated.

There are two logic changes in this PR outside of changes to `configuration.rb`
1. The environment command can no longer reference `root` once in the output since that data no longer lives on the root configuration object.
2.  I found a bug where `Licensed::Git.repository_root` wasn't returning nil as expected when not run from a git repo.  Fixed that and added tests.

### DIFF
--- a/lib/licensed/commands/environment.rb
+++ b/lib/licensed/commands/environment.rb
@@ -23,14 +23,14 @@ module Licensed
             "allowed" => config["allowed"],
             "ignored" => config["ignored"],
             "reviewed" => config["reviewed"],
-            "version_strategy" => self.version_strategy
+            "version_strategy" => self.version_strategy,
+            "root" => config.root
           }
         end
       end
 
       def run(**options)
         super do |report|
-          report["root"] = config.root
           report["git_repo"] = Licensed::Git.git_repo?
         end
       end
--- a/lib/licensed/configuration.rb
+++ b/lib/licensed/configuration.rb
@@ -4,40 +4,39 @@ require "pathname"
 module Licensed
   class AppConfiguration < Hash
     DEFAULT_CACHE_PATH = ".licenses".freeze
-    DEFAULT_CONFIG_FILES = [
-      ".licensed.yml".freeze,
-      ".licensed.yaml".freeze,
-      ".licensed.json".freeze
-    ].freeze
+
+    # Returns the root for a configuration in following order of precendence:
+    # 1. explicitly configured "root" property
+    # 2. a found git repository root
+    # 3. the current directory
+    def self.root_for(configuration)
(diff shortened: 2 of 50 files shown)


jonabc:
The failing `mix` tests look unrelated to this change.  AFAICT the format of `mix.lock` has changed since the `mix` source evaluator was written which is causing the failure.  I'll open a separate issue for that

PR #245, opened 2020-03-07
Add source path expansion using glob patterns

jonabc:
closes https://github.com/github/licensed/issues/239

Adds expansion of `source_path`s set in licensed configuration files.  As noted in the linked issue, this is particularly useful for go projects that commonly have patterns like `<root>/cmd/*` which designate different application/entrypoints into an otherwise shared codebase.  Being able to specify patterns in these situations makes it easier to ensure that dependency are licenses are being evaluated when adding new applications to a repo.


jonabc:
The failed CI jobs look to be [unrelated](https://github.com/haskell/haskell-ide-engine/issues/1680) to this change.

Issue #246, opened 2020-03-14
Go source with modules doesn't work with golang 1.14


jonabc:
With the golang 1.14 release, `go list -e -json -deps` doesn't return a key piece of metadata needed by licensed to determine the folders that it should search for license files.

In go 1.14, the `Module` property for a package in the `go list` output looks like
```
	"Module": {
		"Path": "",
		"Version": ""
	},
```

Whereas in go1.13 it looks like 
```
	"Module": {
		"Path": "",
		"Version": "",
		"Dir": ""
	},
```

In go 1.13, that "Dir" property contains the root of the repository, so if you're looking at an import path like `github.com/stretchr/testify/assert`, the property would have a path like `<repo>/vendor/github.com/stretchy/testify` (assuming that the modules were vendored).  Licensed uses this property as the "search root", to specify that license files should be looked for in all directories between the directory representing the full import path and the directory representing the root of the imported repository.

This has a significant impact as most go projects tend to keep their license texts in the root of the repository, shared between all the packages that the repository exports.  As a result of this change in the `go list` output, licensed ends up not finding license texts for most go packages.

jonabc:
Update: for whatever reason, after a bit of investigation this only seems to affect vendored dependencies 🤷‍♂

PR #247, opened 2020-03-15
Fix go source search_root to work with go 1.14

jonabc:
closes https://github.com/github/licensed/issues/246

This fixes search root calculation to better work for vendored go modules in go 1.14.  The updated calculation checks specifically for a `Module.Dir` property before returning it, and updates the logic for `vendored_path?` to check specify the base of the path that needs to be matched.  In most cases the base path will be the root packages `ImportPath`, but when looking for a search root that value needs to be a directory.  For a vendoring use case the path should be the project root, given by `config.root`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -170,7 +170,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


jonabc:
The CI failures on this PR are from known issues with Cabal.  This PR doesn't touch that source at all though 🤷‍♂

Issue #244, opened 2020-02-28
Allow detection of licenses changed after being `reviewed`


gareth:
This is an open-ended question/feature request because a) I realise this isn't a trivial change and b) I can't even think about how you'd go about implementing it.

At the moment, I can review any dependency that `licensed` doesn't identify as one of my "allowed" licenses, and whitelist it by adding it to the `reviewed:` key. Most commonly this helps where `licensed` tags a dependency as "other", but I'm able to determine that it's just been mislabeled because `licensee` has detected some irrelevant additional text/files in the library.

However, because this dependency is now whitelisted, it's impossible to detect if that license information is later changed for a different license, that I might not allow.

One thing I think *wouldn't* work is changing the `reviewed` syntax to allow `package@version` labels and just compare whether `current_version > reviewed_version`. It would be tedious to force re-review every version bump when nearly all version bumps don't change license information<sup>[citation needed]</sup>.

My first impression is that the "correct" solution for this would require persisting the license cache and actually detecting changes to the license texts themselves - which definitely doesn't sound simple.

I realise that `licensed` isn't supposed to be a strict compliance tool and so I understand this might be `WONTFIX`, but I wanted to open the discussion. If anyone has ideas for an approach then I might be able to look at a PR.

mlinksva:
I agree this would be valuable, and on the correct solution in concept: re-review when license info has changed.

jonabc:
> My first impression is that the "correct" solution for this would require persisting the license cache and actually detecting changes to the license texts themselves - which definitely doesn't sound simple.

Unless I'm misunderstanding, licensed already does this for a different purpose 😄 

When caching licenses, we check the current license text vs the cached license text to determine whether to keep the pre-existing license classification on dependency version changes.
https://github.com/github/licensed/blob/ef02cc78cd4b24b31d18596be0530b086b9879c0/lib/licensed/commands/cache.rb#L49
https://github.com/github/licensed/blob/ef02cc78cd4b24b31d18596be0530b086b9879c0/lib/licensed/dependency_record.rb#L87-L90

Generally this could be done, I'm just unclear on the specific scenario and what should happen when we detect that the license text has changed.

Off the top of my head the solution that would seem to work is to remove the dependency from the `reviewed` list, but given the disconnect between when `licensed cache` is run from `licensed status`, that could get really confusing for users.

gareth:
> Unless I'm misunderstanding, licensed already does this for a different purpose

Oh great!

[...]

PR #248, opened 2020-03-15
Raise an error when a reviewed dependency has changed license text

jonabc:
closes https://github.com/github/licensed/issues/244

Adds a flag to cached metadata during `licensed cache` when
1. there is a cached record for a dependency
2. the cached record's license text does not match the dependency's current license text
3. the dependency is marked `reviewed` in the configuration

If the flag is detected during `licensed status`, it will show an error and `status` will exit with a failure code
```
"license text has changed and needs re-review. if the new text is ok, remove the `review_changed_license` flag from the cached record"
```

I also considered an option of simply unmarking the dependency as `reviewed` in the configuration in this case and letting the current unreviewed license checks catch the scenario, but I'm not sure if that would be confusing for users.

/cc @gareth @mlinksva does this sound right?  I'd be interested in your thoughts on the error text and whether there is a more straightforward solution.  The error feels pretty chatty but I'm not sure how to make it shorter without losing the context of what's wrong and how to fix it.

### DIFF
--- a/lib/licensed/commands/cache.rb
+++ b/lib/licensed/commands/cache.rb
@@ -45,8 +45,15 @@ module Licensed
         filename = app.cache_path.join(source.class.type, "#{dependency.name}.#{DependencyRecord::EXTENSION}")
         cached_record = Licensed::DependencyRecord.read(filename)
         if options[:force] || save_dependency_record?(dependency, cached_record)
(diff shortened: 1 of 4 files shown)


mlinksva:
Sounds right to me. Hopefully rare enough that the chatty context is merited.

PR #249, opened 2020-03-15
update rake for CVE-2020-8130

jonabc:
Updates rake to a newer version for CVE-2020-8130

This is a development dependency and shouldn't impact the shipped version of licensed 👍


jonabc:
Cabal failures are known and being fixed in a separate PR.  Going to ignore them here.

PR #254, opened 2020-03-24
Fix thor dependency

eileencodes:
Rails now requires Thor 1.0 or higher so we need to update the
dependency on thor in licensed. 

This will allow 0.19 or anything higher.


jonabc:
@eileencodes :wave: release 2.9.1 is out with this change https://github.com/github/licensed/releases/tag/2.9.1

PR #261, opened 2020-05-07
Add NuGet source

zarenner:
Add NuGet / dotnet as a supported package source type.

There are a number of open questions remaining since NuGet is somewhat unique in terms of the number of different license mechanisms it has, the most common being fetching licenses from a URL.

FYI this is the first Ruby I've ever written, so I apologize in advance for any atrocities 😃

### DIFF
--- a/lib/licensed/sources/nuget.rb
+++ b/lib/licensed/sources/nuget.rb
@@ -25,7 +25,7 @@ module Licensed
 
         def nuspec_path
           name = @metadata["name"]
-          File.join(self.path, "#{name}.nuspec")
+          File.join(self.path, "#{name.downcase}.nuspec")
         end
 
         def nuspec_contents
--- a/lib/licensed/sources/nuget.rb
+++ b/lib/licensed/sources/nuget.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require "json"
+require "reverse_markdown"
 
 module Licensed
   module Sources
@@ -10,70 +11,14 @@ module Licensed
         "nuget"
       end
 
-      # Match nuspec license expressions like <license type="expression">MIT</license>
-      class NuspecLicenseExpressionMatcher < Licensee::Matchers::Package
-        LICENSE_EXPRESSION_REGEX = /
-          <license\s*type\s*=\s*\"\s*expression\s*\"\s*>\s*([a-z\-0-9\.]+)\s*<\/license>
-        /ix.freeze
-
-        def license_property
-          return @license_property if defined?(@license_property)
-          match = @file.content.match LICENSE_EXPRESSION_REGEX
-          if match && match[1]
-            @license_property = match[1].downcase
-          end
-        end
-      end
-
-      class NuspecFile < Licensee::ProjectFiles::ProjectFile
-        def possible_matchers
-          [NuspecLicenseExpressionMatcher]
-        end
-      end
-
-      # A pseudo-license file where the license is already known (e.g. inferred from a URL)
-      class KnownLicenseFile < Licensee::ProjectFiles::ProjectFile
-        def initialize(license, metadata)
-          super(license.content, metadata)
-          @license = license
-        end
-
-        def license
-          return @license
-        end
-
-        def confidence
-          100
-        end
-      end
-
       class NuGetDependency < Licensed::Dependency
         LICENSE_FILE_REGEX = /<license\s*type\s*=\s*\"\s*file\s*\"\s*>\s*(.*)\s*<\/license>/ix.freeze
         LICENSE_URL_REGEX = /<licenseUrl>\s*(.*)\s*<\/licenseUrl>/ix.freeze
+        PROJECT_URL_REGEX = /<projectUrl>\s*(.*)\s*<\/projectUrl>/ix.freeze
 
-        def license
-          # By default, multiple of the same/similar licenses (e.g. a package with LICENSE.txt, license, and licenseUrl)
-          # will incorrectly result in an "other" license.
-
-          # Treat license expressions as most trustworthy, then local licenses, then remote licenses, then any license files
-          return @license if defined? @license
-          return @license = package_file.license if package_file && package_file.license
-          return @license = nuspec_local_license_file.license if nuspec_local_license_file && nuspec_local_license_file.license
-          return @license = nuspec_remote_license_file.license if nuspec_remote_license_file && nuspec_remote_license_file.license
-          super
-        end
-
-        def package_name
-          return unless @metadata
-          @metadata["name"]
-        end
-
-        def package_file
-          @package_file ||= NuspecFile.new(nuspec_contents, nuspec_path)
-        end
-
-        def project_files
-          @project_files ||= [license_files, readme_file, package_file, nuspec_local_license_file, nuspec_remote_license_file].flatten.compact
+        def initialize(name:, version:, path:, search_root: nil, metadata: {}, errors: [])
+          super(name: name, version: version, path: path, search_root: search_root, metadata: metadata, errors: errors)
+          @metadata["homepage"] = project_url if project_url
         end
 
         def nuspec_path
@@ -86,19 +31,40 @@ module Licensed
           @nuspec_contents ||= File.read(nuspec_path)
         end
 
+        def project_url
+          return unless nuspec_contents
+          @project_url ||= begin
+            match = nuspec_contents.match PROJECT_URL_REGEX
+            match[1] if match && match[1]
+          end
+        end
+
+        def project_files
+          @nuget_project_files ||= begin
+            files = [super(), nuspec_local_license_file].flatten.compact
+
+            # Only download licenseUrl if no recognized license was found locally
+            if files.none? { |file| file.license && file.license.key != "other" }
+              files.push(nuspec_remote_license_file)
+            end
+            files.compact
+          end
+        end
+
         # Look for a <license type="file"> element in the nuspec that points to an
         # on-disk license file (which licensee may not find due to a non-standard filename)
         def nuspec_local_license_file
           return @nuspec_local_license_file if defined?(@nuspec_local_license_file)
           return unless nuspec_contents
 
-          match = @nuspec_contents.match LICENSE_FILE_REGEX
-          if match && match[1]
-            license_path = File.join(File.dirname(nuspec_path), match[1])
-            return unless File.exist?(license_path)
-            license_data = File.read(license_path)
-            @nuspec_local_license_file = Licensee::ProjectFiles::LicenseFile.new(license_data, license_path)
-          end
+          match = nuspec_contents.match LICENSE_FILE_REGEX
+          return unless match && match[1]
+
+          license_path = File.join(File.dirname(nuspec_path), match[1])
+          return unless File.exist?(license_path)
+
+          license_data = File.read(license_path)
+          @nuspec_local_license_file = Licensee::ProjectFiles::LicenseFile.new(license_data, license_path)
         end
 
         # Look for a <licenseUrl> element in the nuspec that either is known to contain a license identifier
@@ -108,59 +74,69 @@ module Licensed
           return unless nuspec_contents
 
           match = nuspec_contents.match LICENSE_URL_REGEX
-          if match && match[1]
-            url = original_url = match[1]
+          return unless match && match[1]
 
-            # Skip downloading if the URL contains a license identifier itself
-            if known_license = inspect_url_for_license(url)
-              return @nuspec_remote_license_file = KnownLicenseFile.new(known_license, { uri: url })
+          # Attempt to fetch the license content
+          license_content = self.class.retrieve_license(match[1])
+          @nuspec_remote_license_file = Licensee::ProjectFiles::LicenseFile.new(license_content, { uri: match[1] }) if license_content
+        end
+
+        class << self
+          def strip_html(html)
+            return unless html
+
+            if html.downcase.include?("<html")
+              ReverseMarkdown.convert(html, unknown_tags: :bypass)
+            else
+              html
             end
+          end
 
-            unless ignored_url?(url)
-              # Transform URLs that are known to return HTML but have a corresponding text-based URL
-              url = get_text_content_url(url)
+          def ignored_url?(url)
+            # Many Microsoft packages that now use <license> use this for <licenseUrl>
+            # No need to fetch this page - it just contains NuGet documentation
+            url == "https://aka.ms/deprecateLicenseUrl"
+          end
 
-              # Attempt to fetch the license content
-              license_data = retrieve_license(url)
-              unless license_data.nil?
-                @nuspec_remote_license_file = Licensee::ProjectFiles::LicenseFile.new(license_data, { uri: original_url })
+          def text_content_url(url)
+            # Convert github file URLs to raw URLs
+            return url unless match = url.match(/https?:\/\/(?:www\.)?github.com\/([^\/]+)\/([^\/]+)\/blob\/(.*)/i)
+            "https://github.com/#{match[1]}/#{match[2]}/raw/#{match[3]}"
+          end
+
+          def retrieve_license(url)
+            return unless url
+            return if ignored_url?(url)
+
+            # Transform URLs that are known to return HTML but have a corresponding text-based URL
+            text_url = text_content_url(url)
+
+            raw_content = fetch_content(text_url)
+            strip_html(raw_content)
+          end
+
+          def fetch_content(url, redirect_limit = 5)
+            url = URI.parse(url) if url.instance_of? String
+            return @response_by_url[url] if (@response_by_url ||= {}).key?(url)
+            return if redirect_limit == 0
+
+            begin
+              response = Net::HTTP.get_response(url)
+              case response
+              when Net::HTTPSuccess     then
+                @response_by_url[url] = response.body
+              when Net::HTTPRedirection then
+                redirect_url = URI.parse(response["location"])
+                if redirect_url.relative?
+                  redirect_url = url + redirect_url
+                end
+                # The redirect might be to a URL that requires transformation, i.e. a github file
+                redirect_url = text_content_url(redirect_url.to_s)
+                @response_by_url[url] = fetch_content(redirect_url, redirect_limit - 1)
               end
+            rescue => error
+              # Host might no longer exist or some other error, ignore
             end
-          end
-        end
-
-        def get_text_content_url(url)
-          # Convert github file URLs to raw URLs
-          if match = url.match(/https?:\/\/(?:www\.)?github.com\/([^\/]+)\/([^\/]+)\/blob\/(.*)/i)
-            url = "https://github.com/#{match[1]}/#{match[2]}/raw/#{match[3]}"
-          else
-            url
-          end
-        end
-
-        def inspect_url_for_license(url)
-          if match = url.match(/https?:\/\/licenses.nuget.org\/(.*)/i)
-            Licensee::License.find(match[1])
-          elsif match = url.match(/https?:\/\/opensource.org\/licenses\/(.*)/i)
-            Licensee::License.find(match[1])
-          elsif match = url.match(/https?:\/\/(?:www\.)?apache.org\/licenses\/(.*?)(?:\.html|\.txt)?$/i)
-            Licensee::License.find(match[1].gsub("LICENSE", "Apache"))
-          end
-        end
-
-        def ignored_url?(url)
-          url == "https://aka.ms/deprecateLicenseUrl"
-        end
-
-        def retrieve_license(url, redirect_limit = 2)
-          return if redirect_limit == 0
-          begin
-            response = Net::HTTP.get_response(URI(url))
-            case response
-            when Net::HTTPSuccess     then response.body
-            when Net::HTTPRedirection then retrieve_license(response["location"], redirect_limit - 1)
-            end
-          rescue
           end
         end
       end
@@ -173,6 +149,20 @@ module Licensed
         config.dig("nuget", "obj_root") || config.pwd
       end
 
+      def dump_projects?
+        config.dig("nuget", "projects", "dump")
+      end
+
+      def exclude_projects
+        config.dig("nuget", "projects", "exclude") || []
+      end
+
+      def excluded_project?(project_name)
+        exclude_projects.any? do |pattern|
+          File.fnmatch?(pattern, project_name, File::FNM_PATHNAME | File::FNM_CASEFOLD)
+        end
+      end
+
       def nuget_config_exists(root)
         # Multiple supported casings: https://github.com/NuGet/Home/issues/1427
         File.exist?(File.join(root, "nuget.config")) || File.exist?(File.join(root, "NuGet.config")) || File.exist?(File.join(root, "NuGet.Config"))
@@ -180,14 +170,20 @@ module Licensed
 
       def enumerate_dependencies
         packages.map do |key, package|
+          metadata = {
+            "type" => NuGet.type,
+            "name" => package["name"]
+          }
+
+          # Emit the names of the projects that consume a particular dependency.
+          # Useful for determining projects to exclude.
+          metadata["projects"] = package["projects"].to_a.sort if dump_projects?
+
           NuGetDependency.new(
-            name: "#{package["name"]} #{package["version"]}",
+            name: "#{package["name"]}-#{package["version"]}",
             version: package["version"],
             path: package["path"],
-            metadata: {
-              "type"     => NuGet.type,
-              "name"     => package["name"]
-            }
+            metadata: metadata
           )
         end
       end
@@ -208,16 +204,24 @@ module Licensed
       def gather_packages(project_assets_file)
         json = JSON.parse(File.read(project_assets_file))
         nuget_packages_dir = json["project"]["restore"]["packagesPath"]
-        json["targets"].keys.map do |target|
-          json["targets"][target].keys.map do |reference|
-            next if @packages_by_id.has_key?(reference)
-            next unless json["targets"][target][reference]["type"] == "package"
-            package_id_parts = reference.partition("/")
-            path = File.join(nuget_packages_dir, json["libraries"][reference]["path"])
-            @packages_by_id[reference] = {
-              "name"    => package_id_parts[0],
-              "version" => package_id_parts[-1],
-              "path"    => path }
+        project_name = json["project"]["restore"]["projectName"]
+        return if excluded_project?(project_name)
+
+        json["targets"].map do |target_key, target|
+          target.map do |reference_key, reference|
+            next unless reference["type"] == "package"
+            package_id_parts = reference_key.partition("/")
+            path = File.join(nuget_packages_dir, json["libraries"][reference_key]["path"])
+            if @packages_by_id.key?(reference)
+              @packages_by_id[reference_key]["projects"].add(project_name)
+            else
+              @packages_by_id[reference_key] = {
+                "name"     => package_id_parts[0],
+                "version"  => package_id_parts[-1],
+                "path"     => path,
+                "projects" => Set.new([project_name])
+              }
+            end
           end
         end
       end
--- a/lib/licensed/sources/nuget.rb
+++ b/lib/licensed/sources/nuget.rb
@@ -1,0 +1,226 @@
+# frozen_string_literal: true
+require "json"
+
+module Licensed
+  module Sources
+    # Only supports ProjectReference (project.assets.json) style restore used in .NET Core.
+    # Does not currently support packages.config style restore.
+    class NuGet < Source
+      def self.type
+        "nuget"
+      end
+
+      # Match nuspec license expressions like <license type="expression">MIT</license>
+      class NuspecLicenseExpressionMatcher < Licensee::Matchers::Package
+        LICENSE_EXPRESSION_REGEX = /
+          <license\s*type\s*=\s*\"\s*expression\s*\"\s*>\s*([a-z\-0-9\.]+)\s*<\/license>
+        /ix.freeze
+
+        def license_property
+          return @license_property if defined?(@license_property)
+          match = @file.content.match LICENSE_EXPRESSION_REGEX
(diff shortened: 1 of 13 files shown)


lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,226 @@
...
+      def packages
+        return @packages_by_id if defined?(@packages_by_id)
+        @packages_by_id = Hash.new
+        Dir.glob(File.join(nuget_obj_root, "**/project.assets.json")).map do |file|
+          gather_packages(file)
+        end
+        @packages_by_id
+      end
+
+      def gather_packages(project_assets_file)
```
zarenner, line 208: Is it expected to pull in transitive dependencies? I assumed so, but just wanted to check since it results in lots of .NET Core implementation detail packages that have to be ignored.

zarenner:
A large number of Microsoft packages that any ASP.NET Core project will likely use have a specific, not SPDX recognized license. Is there any way to have this license show up as something different than other so that they don’t have to manually be reviewed? It looks like anything not in choosealicense is consider an invalid license.

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,226 @@
...
+
+      def nuget_config_exists(root)
+        # Multiple supported casings: https://github.com/NuGet/Home/issues/1427
+        File.exist?(File.join(root, "nuget.config")) || File.exist?(File.join(root, "NuGet.config")) || File.exist?(File.join(root, "NuGet.Config"))
+      end
+
+      def enumerate_dependencies
+        packages.map do |key, package|
+          NuGetDependency.new(
+            name: "#{package["name"]} #{package["version"]}",
```
zarenner, line 184: How should multiple versions of the same dependency be handled? 

It looks like multiple dependencies with the same name (not metadata.name) result in only one file being emitted into the cache which makes status unhappy, so I made the dependency name / outputted file be “{name} {version}“, but no other source appears to do so.

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,226 @@
...
+            if known_license = inspect_url_for_license(url)
+              return @nuspec_remote_license_file = KnownLicenseFile.new(known_license, { uri: url })
+            end
+
+            unless ignored_url?(url)
+              # Transform URLs that are known to return HTML but have a corresponding text-based URL
+              url = get_text_content_url(url)
+
+              # Attempt to fetch the license content
+              license_data = retrieve_license(url)
```
zarenner, line 124: There’s still a good number of licenses that will come back as HTML. Should these be ignored? Included as HTML? Parsed to text?

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,226 @@
...
+
+        def license
+          # By default, multiple of the same/similar licenses (e.g. a package with LICENSE.txt, license, and licenseUrl)
+          # will incorrectly result in an "other" license.
+
+          # Treat license expressions as most trustworthy, then local licenses, then remote licenses, then any license files
+          return @license if defined? @license
+          return @license = package_file.license if package_file && package_file.license
+          return @license = nuspec_local_license_file.license if nuspec_local_license_file && nuspec_local_license_file.license
+          return @license = nuspec_remote_license_file.license if nuspec_remote_license_file && nuspec_remote_license_file.license
```
zarenner, line 62: Thoughts on fetching `licenseUrl` if other licenses were already found from a file or license expression?

On one hand, it seems wise to avoid doing so, on the other hand the online license may be different than the file.

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,226 @@
...
+          return @nuspec_remote_license_file if defined?(@nuspec_remote_license_file)
+          return unless nuspec_contents
+
+          match = nuspec_contents.match LICENSE_URL_REGEX
+          if match && match[1]
+            url = original_url = match[1]
+
+            # Skip downloading if the URL contains a license identifier itself
+            if known_license = inspect_url_for_license(url)
+              return @nuspec_remote_license_file = KnownLicenseFile.new(known_license, { uri: url })
```
zarenner, line 116: The source is listed as `licenseUrl` though the license text is auto-generated (rather than downloaded from that URI) - just checking that this is ok. An alternative is to have the source be the "Auto-generated" text.

lib/licensed/sources/nuget.rb
jonabc: It looks like licensee already handles [html files](https://github.com/licensee/licensee/blob/4a9133ed3abf46fbe5a30ca1f1846484807f6eb2/lib/licensee/content_helper.rb#L295-L301) though you might need to jump through some hoops to get that working with the license data downloaded from a url.

Alternatively, since [reverse_markdown](https://github.com/xijo/reverse_markdown) is already installed transitively through licensee it may make sense to add it to `licensed.gemspec` directly and call it manually to make it easier.

Either way, as long as we can give licensee the best chance to parse and classify the content, 👍

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,226 @@
...
+      end
+
+      class NuspecFile < Licensee::ProjectFiles::ProjectFile
+        def possible_matchers
+          [NuspecLicenseExpressionMatcher]
+        end
+      end
+
+      # A pseudo-license file where the license is already known (e.g. inferred from a URL)
+      class KnownLicenseFile < Licensee::ProjectFiles::ProjectFile
```
jonabc, line 35: 👍 this sounds helpful.  since it's generic and not tied to this class, can be moved to its own file like `lib/licensed/project_files/known_license_file.rb`?

 or maybe licensee would be ok with having this class as an official licensee project file type?

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,226 @@
...
+          elsif match = url.match(/https?:\/\/(?:www\.)?apache.org\/licenses\/(.*?)(?:\.html|\.txt)?$/i)
+            Licensee::License.find(match[1].gsub("LICENSE", "Apache"))
+          end
+        end
+
+        def ignored_url?(url)
+          url == "https://aka.ms/deprecateLicenseUrl"
+        end
+
+        def retrieve_license(url, redirect_limit = 2)
```
jonabc, line 155: what are the chances that two dependencies point at the same license url here?  just wondering if the output from this function should be cached/memoized

zarenner:
> ❤️ thanks for opening this!
> 
> It's been awhile since I've done any .net development. Is this implementation specific to any specific .net version or is it a general solution that works for anything that uses nuget?

This implementation works for all .NET Core projects, and newer .NET Framework projects that use [PackageReference](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files), where nuget resolves the package graph during restore and produces files indicating where the packages were downloaded to.

It will not work for older .NET Framework projects that use the `packages.config` mechanism. Implementing support for `packages.config` here is quite difficult as it's harder to find the packages on disk without parsing MSBuild.

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,226 @@
...
+            @license_property = match[1].downcase
+          end
+        end
+      end
+
+      class NuspecFile < Licensee::ProjectFiles::ProjectFile
+        def possible_matchers
+          [NuspecLicenseExpressionMatcher]
+        end
+      end
```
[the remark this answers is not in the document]
zarenner: I'll open a PR there - I wasn't sure if they'd accept "partial" support for NuGet licenses, but happy to try.

lib/licensed/sources/nuget.rb
zarenner: Great point - that's in fact very common and definitely worth doing.

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,226 @@
...
+        nuget_config_exists(config.pwd)
+      end
+
+      def nuget_obj_root
+        config.dig("nuget", "obj_root") || config.pwd
+      end
+
+      def nuget_config_exists(root)
+        # Multiple supported casings: https://github.com/NuGet/Home/issues/1427
+        File.exist?(File.join(root, "nuget.config")) || File.exist?(File.join(root, "NuGet.config")) || File.exist?(File.join(root, "NuGet.Config"))
```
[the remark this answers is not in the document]
zarenner: Only on a case-insensitive filesystem (like the default in macOS) 😃 
Linux test: https://repl.it/repls/OldlaceMessyHexagons

lib/licensed/sources/nuget.rb
zarenner: Sort of - https://docs.microsoft.com/en-us/dotnet/core/packages#packages lists the "key" root packages. I don't know how stable that list is nor if they're all licensed exactly the same as the .NET Core runtime, so I avoided using it. I think that it's probably best to leave them for now and document that `ignored` can be used to ignore e.g. `runtime.*`, `System.*` if desired.

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,226 @@
...
+        Dir.glob(File.join(nuget_obj_root, "**/project.assets.json")).map do |file|
+          gather_packages(file)
+        end
+        @packages_by_id
+      end
+
+      def gather_packages(project_assets_file)
+        json = JSON.parse(File.read(project_assets_file))
+        nuget_packages_dir = json["project"]["restore"]["packagesPath"]
+        json["targets"].keys.map do |target|
```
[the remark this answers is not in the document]
zarenner: Thanks, learning Ruby as I go 😄

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,226 @@
...
+        end
+      end
+
+      class NuGetDependency < Licensed::Dependency
+        LICENSE_FILE_REGEX = /<license\s*type\s*=\s*\"\s*file\s*\"\s*>\s*(.*)\s*<\/license>/ix.freeze
+        LICENSE_URL_REGEX = /<licenseUrl>\s*(.*)\s*<\/licenseUrl>/ix.freeze
+
+        def license
+          # By default, multiple of the same/similar licenses (e.g. a package with LICENSE.txt, license, and licenseUrl)
+          # will incorrectly result in an "other" license.
```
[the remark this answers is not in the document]
zarenner: Yeah, you must be right; one of them must have been classified as `other`.

I think the general issue is that since there's almost always a `licenseUrl` specified, we'll get back a lot of HTML or otherwise low-quality text that's likely to be classified as `other`.

I agree overriding the licensee behavior feels wrong, but it also feels wrong to me to have a significant number of packages that are using the new nuget `license` SPDX expressions get classified as `other` just because they decided to also include a URL to the HTML version of the license.

I'll see how implementing HTML -> text affects this on my sample package set, but I'm guessing it still won't be good enough to avoid `other` classifications.

I'll open an issue on licensee since it seems like `Project.license` should take into account the confidence especially if one is 100 and the other is near-zero.
An alternative if licensee doesn't accept a change here might be to just ignore HTML licenses.

lib/licensed/sources/nuget.rb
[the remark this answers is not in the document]
zarenner: https://github.com/licensee/licensee/pull/432

lib/licensed/sources/nuget.rb
jonabc: Is there anything to stop someone from building a package like `System.MyPackage`?  Conventions maybe?

lib/licensed/sources/nuget.rb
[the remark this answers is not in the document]
jonabc: ahhhhh TIL

lib/licensed/sources/nuget.rb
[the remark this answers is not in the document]
jonabc: ah gotcha, would you like me to give a deeper look and suggestions/comments for ruby style?

lib/licensed/sources/nuget.rb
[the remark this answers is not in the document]
zarenner: At least for my test cases, using ReverseMarkdown seems to work pretty well and I don't think this is a major issue.

lib/licensed/sources/nuget.rb
zarenner: Purely conventional. Nothing to actually stop someone from doing that AFAIK.

lib/licensed/sources/nuget.rb
zarenner: Added reverse_markdown, appears to work better than I expected!

lib/licensed/sources/nuget.rb
zarenner: I no longer override `license`, but I still have custom logic in `project_files` to only download licenseUrl if neither licensee nor the local file search found a recognized license.

With the change I did in `licensee`, it's no longer possible to differentiate if `licensee` actually matched on `<license>` or `<licenseUrl>`. I guess I could make another change in licensee to add that back in, but to be honest I'd rather not if not necessary.

Thoughts?

lib/licensed/sources/nuget.rb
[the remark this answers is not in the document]
zarenner: Actually, I spoke too soon. ReverseMarkdown does a great job at making the text readable, but it still results in lots of 'other' classifications. If I were to always download licenseUrl (regardless of if it or <license> matched), most packages with html licenses then become other.

lib/licensed/sources/nuget.rb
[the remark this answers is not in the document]
zarenner: Definitely would appreciate that, thanks!

lib/licensed/sources/nuget.rb
mlinksva: That custom logic sounds right to me.

lib/licensed/sources/nuget.rb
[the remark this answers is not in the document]
jonabc: 🤔 is this on the test fixture, or a project you're using this with?  if it's on something different, can you put together an example project you can share publicly that I can take a look at?

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,230 @@
...
+          return unless nuspec_path
+          @nuspec_contents ||= File.read(nuspec_path)
+        end
+
+        def project_url
+          return unless nuspec_contents
+          @project_url ||= begin
+            match = nuspec_contents.match PROJECT_URL_REGEX
+            match[1] if match && match[1]
+          end
```
jonabc, line 39: The caching here will fail if the evaluated block returns nil.  If that happens, follow up calls to this expand to `@project_url = nil || begin`, which will re-execute the block.

This can be fixed by checking whether `@project_url` has been defined instead of relying on `||=` 👍 

```suggestion
          return @project_url if defined?(@project_url)
          @project_url = begin
            match = nuspec_contents.match PROJECT_URL_REGEX
            match[1] if match && match[1]
          end
```

lib/licensed/sources/nuget.rb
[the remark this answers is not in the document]
zarenner: Any package that uses https://licenses.nuget.org is a good example, since it doesn't return the raw license text; it returns HTML with "SPDX identifier" and "License text" sections. e.g. https://licenses.nuget.org/MIT

ReverseMarkdown'ing that page results in fairly readable license text, but is not identified as MIT because of `## SPDX identifier\n\nMIT\n\n## License text\n\n`.

Special-casing licenses.nuget.org is an option, of course. But as I mentioned somewhere - if licenses.nuget.org is used, the package author very likely used `<license>` anyways. Let me see if I can find other examples.

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,230 @@
...
+        end
+
+        def project_files
+          @nuget_project_files ||= begin
+            files = [super(), nuspec_local_license_file].flatten.compact
```
jonabc, line 44: If a local license file referenced from the nuspec has a standard filename, e.g. `LICENSE`, this will recognize that file twice.  WDYT about only adding `nuspec_local_license_file` if `files` is either empty or only contains `Licensee::ProjectFiles::PackageManagerFile` objects?

lib/licensed/sources/nuget.rb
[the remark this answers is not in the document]
zarenner: http://www.apache.org/licenses/LICENSE-2.0.html is another example. `licensee` recognizes it, but a ReverseMarkdown of the actual page is certainly not the license text:
```
text: "Toggle navigation\n\n- \n\n- [News](/index.html#news)\n- [About&nbsp;](#)\n
    \ - [Overview](/foundation)\n  - [Process](/foundation/how-it-works.html)\n  -
    [Governance](/foundation/governance/)\n  - [The Apache Way](/theapacheway/index.html)\n
    \ - [Membership](/foundation/governance/members.html)\n  - [Community](https://community.apache.org/)\n
...
```
This is a case where the license text is effectively useless and will need to manually be removed, in addition to the incorrect 'other' classification.

Another case is http://www.microsoft.com/web/webpi/eula/net_library_eula_enu.htm. While it's not a recognized license (although I wish it were since it's so common), it would likely have a similar issue since ReverseMarkdown converts links to markdown-style links, leaves some nbsp; around, etc. In this case, ReverseMarkdown produces good enough text to read, but unlikely that it'd be good enough to detect.

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,230 @@
...
+
+        def project_url
+          return unless nuspec_contents
+          @project_url ||= begin
+            match = nuspec_contents.match PROJECT_URL_REGEX
+            match[1] if match && match[1]
+          end
+        end
+
+        def project_files
```
jonabc, line 42: it would duplicate a bit of logic, but if you find a recognized license in a `Licensee::ProjectFiles::PackageManagerFile` and its contents match `Licensee::Matchers::Nuget::LICENSE_URL_REGEX` then you could add a license file like
```
# strip copyright clauses and any extra newlines
text = file.license.text
                   .lines
                   .reject { |l| l =~ Licensee::Matchers::Copyright::REGEX }
                   .join
                   .gsub(/\n\n\n/, "\n\n")

Licensee::ProjectFiles::LicenseFile.new(text, { uri: <matched url> })
```

The text generation was copied from [dependency.rb](https://github.com/github/licensed/blob/b5e8cb9b7257d7102a656cb6e242ab14b3c3287c/lib/licensed/dependency.rb#L140-L158).

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,230 @@
...
+
+        class << self
+          def strip_html(html)
+            return unless html
+
+            if html.downcase.include?("<html")
+              ReverseMarkdown.convert(html, unknown_tags: :bypass)
+            else
+              html
+            end
```
jonabc, line 92: ruby idiom is to use a guard clause 👍 
```suggestion
            return html unless html.downcase.include?("<html")
            ReverseMarkdown.convert(html, unknown_tags: :bypass)
```

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,230 @@
...
+              when Net::HTTPRedirection then
+                redirect_url = URI.parse(response["location"])
+                if redirect_url.relative?
+                  redirect_url = url + redirect_url
+                end
+                # The redirect might be to a URL that requires transformation, i.e. a github file
+                redirect_url = text_content_url(redirect_url.to_s)
+                @response_by_url[url] = fetch_content(redirect_url, redirect_limit - 1)
+              end
+            rescue => error
```
jonabc, line 137: if the error is unused, you don't need to capture it to a variable

```suggestion
            rescue
```

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,230 @@
...
+
+      def dump_projects?
+        config.dig("nuget", "projects", "dump")
+      end
+
+      def exclude_projects
+        config.dig("nuget", "projects", "exclude") || []
+      end
+
+      def excluded_project?(project_name)
```
jonabc, line 160: can you provide a comment here on what the usage here is?  Is this so you can set an exclude for something like `System.*`?  if so, why not specify that in the `ignore.nuget` configuration list instead of creating a new configuration value?

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,230 @@
...
+      end
+
+      def enumerate_dependencies
+        packages.map do |key, package|
+          metadata = {
+            "type" => NuGet.type,
+            "name" => package["name"]
+          }
+
+          # Emit the names of the projects that consume a particular dependency.
```
jonabc, line 178: 👍 being able to see how a package is included in a project sounds really nice, but I don't understand this comment - how does this tie back to excluding projects?

I'm a little concerned that exclusions are applied to not evaluate nested dependencies, which would be problematic if the excluded dependency was updated to bring in a new dependency that should not be excluded.  

Dependency sources should always evaluate the full dependency tree, with the `ignore` setting used to skip further operations (cache, status, list) on individual dependencies.

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,230 @@
...
+      def packages
+        return @packages_by_id if defined?(@packages_by_id)
+        @packages_by_id = Hash.new
+        Dir.glob(File.join(nuget_obj_root, "**/project.assets.json")).map do |file|
+          gather_packages(file)
+        end
+        @packages_by_id
+      end
+
+      def gather_packages(project_assets_file)
```
jonabc, line 204: A few ruby idioms on this one 👍 
1. It's preferable to pass state into functions rather than mutate an instance variable.
2. `each` is preferable to `map` in this case since you're not using the returned array from `map` 
3. You can clean this up by using `each_with_object` so you don't need to predefine `@package_by_id`

```suggestion
      def packages
        return @packages_by_id if defined?(@packages_by_id)
        search_path = File.join(nuget_obj_root, "**/project.assets.json")
        @packages_by_id = Dir.glob(search_path).each_with_object({}) do |file, packages|
          gather_packages(file, packages)
        end
      end

      def gather_packages(project_assets_file, packages)
```

With further updates to `gather_packages` to use the passed in hash rather than the instance variable.

lib/licensed/sources/nuget.rb
zarenner: In .NET Core, packages are now specified on a per-project basis. That is, there is no packages.config/project.json/etc for the whole solution (repo).

This would make it possible to exclude the packages used by any [test | build-time | non-shipping] projects.
e.g. 
```
projects:
    exclude:
      - "*.Mocks"
      - "*.Tests"
      - Tools.CodeAnalyzers
      - InternalNonShippingService
```

where `Foo.Tests.csproj` is a project that consumes the package `Moq`, which is only used for tests and is never shipped.

I'm happy to take this out if it's against licensed's principles, but for our repository it would result in ~100 packages that would need to be ignored, and make our `.licensed.yml` somewhat unmanageable.

lib/licensed/sources/nuget.rb
zarenner: See above comment. This isn't about transitive dependencies - it's about excluding [test | non-shipping] projects from pulling in root dependencies.

lib/licensed/sources/nuget.rb
zarenner: Makes sense, thanks!

lib/licensed/sources/nuget.rb
zarenner: Thanks, was not aware of the use case for `each_with_object`, now it makes sense. Will fix up gather_packages.

lib/licensed/sources/nuget.rb
zarenner: It feels a bit weird to me since it means we're potentially ignoring the actual license file according to NuGet and choosing a different one... but that's admittedly pretty unlikely.

Any reason not to just check that `files` doesn't already contain that filepath?

lib/licensed/sources/nuget.rb
zarenner: I assume you mean checking that the contents of the `PackageManagerFile` itself match against `LICENSE_URL_REGEX`? 

If so, it's possible that the `PackageManagerFile` license actually came from `Licensee::Matchers::Nuget::LICENSE_REGEX` (`license` expression) and is potentially a different license than `licenseUrl`, so I think you'd also want to check that the package file contents *doesn't* match against `LICENSE_REGEX`.

That would definitely be better though than the current auto-generated source. I was wondering if this sort of change might belong in licensee, but I think it would require some non-trivial refactoring of `PackageManagerFile` which is why I didn't.

If you're ok without this change for now, it might make sense to do this as a separate PR (which I'd be happy to do).

lib/licensed/sources/nuget.rb
zarenner: Made a change to simply check that `files` doesn't already have a file with that name, let me know what you think of that.

lib/licensed/sources/nuget.rb
jonabc: > I assume you mean checking that the contents of the PackageManagerFile itself match against LICENSE_URL_REGEX?

Yep 👍 

> If so, it's possible that the PackageManagerFile license actually came from Licensee::Matchers::Nuget::LICENSE_REGEX (license expression) and is potentially a different license than licenseUrl,

Oh, does the licensee change not look for both of them and report other if they don't match?

> If you're ok without this change for now, it might make sense to do this as a separate PR (which I'd be happy to do).

yeah for sure its fine not to block this one.  the impact being that you might get a bunch of minimal changes to your cached data if the followup PR is put into a future release

lib/licensed/sources/nuget.rb
jonabc: just to make sure I understand - the purpose of this is to exclude projects from the repo being evaluated? 

Put another way - In the example above, all of the projects that are intended to match the listed exclusions exist in the repo which is running `licensed`?

If that's the case, you don't need to run licensed from the root of your repo.  licensed allows you to specify multiple evaluation scopes, called [`apps`](https://github.com/github/licensed/blob/master/docs/configuration.md#specifying-multiple-apps) in the documentation.  I would recommend specifying multiple apps for the projects you're looking to enumerate.  You can specify [glob patterns](https://github.com/github/licensed/blob/master/docs/configuration.md#source-path-glob-patterns) for an app source path which will expand to multiple apps when run in a repo.

Does that help your use case?

lib/licensed/sources/nuget.rb
zarenner: > Oh, does the licensee change not look for both of them and report other if they don't match?

No. Perhaps it should have, but I think I saw that others don't do that (e.g. gemspec matcher doesn't ensure `license` and `licenses` don't conflict), and `licenseUrl` is [officially deprecated](https://docs.microsoft.com/en-us/nuget/reference/nuspec#licenseurl) so it seemed reasonable to ignore `licenseUrl` if `license` is set.

And in the case where `licenseUrl` is also set but not a recognized URL, you definitely wouldn't want to report `other`.

lib/licensed/sources/nuget.rb
zarenner: Correct and correct.

I'm not sure it completely does. I think it's possible to implement, but for large solutions it doesn't really seem reasonable to me.

For context:
We don't even look for "source" project files e.g. `Foo.csproj` today; we look directly for `project.assets.json` files which are emitted after `dotnet restore` runs. Mapping directly from `Foo.csproj` -> `project.assets.json` filepath isn't possible without invoking msbuild. Mapping back from a `project.assets.json` -> `Foo.csproj` (or project name `Foo`) is thankfully easy, however.

Which means that we would need to first build up the mapping before being able to find dependencies for a project. Presumably we'd just do this when evaluating the first one and then cache it for the subsequent projects?

In terms of if it would be useful: For our codebase with hundreds of projects (even with most of them having naming conventions), it seems much much safer to exclude rather than include projects. I would honestly have no clue if I successfully included all of our shipping projects using just include patterns.

lib/licensed/sources/nuget.rb
zarenner: Per our discussion, removed this and at least for now will only support targeting 1 project at a time, where the `source_path` must be the directory containing the `project.assets.json` file.

lib/licensed/sources/nuget.rb
zarenner: Per our discussion, removed this and at least for now will only support targeting 1 project at a time, where the `source_path` must be the directory containing the `project.assets.json` file.

lib/licensed/sources/nuget.rb
zarenner: Resolving per our discussion since this approach seems reasonable.

lib/licensed/sources/nuget.rb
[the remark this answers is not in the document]
zarenner: Per our discussion, I am going to leave the current behavior of downloading the license text only when no other match (package file or license file) is found. After this PR is merged though I'll file an issue to consider revisiting this, either with or without additional changes to `licensee`.

lib/licensed/sources/nuget.rb
zarenner: Per discussion, will leave as is and file an issue to consider for the future.

jonabc:
Ignore the failing cabal tests, it's a common occurrence that something in a new cabal release breaks the setup script 😢 .  I'll get around to fixing it when I have some free time.

jonabc:
I've updated the test GH Action workflow to fix the failing cabal builds, they should be back to ✅ after updating from master.

lib/licensed/sources/nuget.rb
```diff
@@ -0,0 +1,206 @@
...
+              version: version,
+              path: path,
+              metadata: {
+                "type" => NuGet.type,
+                "name" => name
+              }
+            )
+          end
+        end.values
+      end
```
jonabc, line 203: nit: this would be easier to read and could avoid the dangling `.values` if this was broken up into separate methods that (1) gathered metadata and (2) mapped metadata to `NugetDependency`. 

not blocking, 👍 to clean this up later so this source can be released.

PR #277, opened 2020-05-29
Add command to output `NOTICE` file

jonabc:
This add a new command `licensed notices`, which outputs a `NOTICE` file for each configured application.  I couldn't find a standard format for the file, so for now it looks like the following, and I'd welcome feedback if there is a preferred format.

```
THIRD PARTY NOTICES

*****
dependency@version

license text
-----
license text
-----
notice text
-----
notice text

*****
dependency@version
.....
```

Most of the work happens in the reporter, though it could arguably be done in the command as well  🤷 

/cc @mlinksva @jeffmcaffer any opinions on the NOTICE file format?


jonabc:
> any opinions on the NOTICE file format?

@mlinksva @jeffmcaffer ping :wave:

jeffmcaffer:
Not from me per se. As long as it meets legal terms and doesn't look completely horrible (it doesn't) then box checked!

PR #273, opened 2020-05-25
Lookup packages in the NugetFallbackFolders

paveliak:
Some nuget packages may be located in the `NugetFallbackFolders`. We need to scan it too when searching for a package or `licensed cache` would throw exception like this
```
paveliak@Pavels-MacBook-Pro src % licensed cache
Caching dependency records for LightRail.Actions
  nuget
Traceback (most recent call last):
	42: from /usr/local/bin/licensed:23:in `<main>'
	41: from /usr/local/bin/licensed:23:in `load'
	40: from /Library/Ruby/Gems/2.6.0/gems/licensed-2.10.0/exe/licensed:5:in `<top (required)>'
	39: from /Library/Ruby/Gems/2.6.0/gems/thor-1.0.1/lib/thor/base.rb:485:in `start'
	38: from /Library/Ruby/Gems/2.6.0/gems/thor-1.0.1/lib/thor.rb:392:in `dispatch'
	37: from /Library/Ruby/Gems/2.6.0/gems/thor-1.0.1/lib/thor/invocation.rb:127:in `invoke_command'
	36: from /Library/Ruby/Gems/2.6.0/gems/thor-1.0.1/lib/thor/command.rb:27:in `run'
	35: from /Library/Ruby/Gems/2.6.0/gems/licensed-2.10.0/lib/licensed/cli.rb:14:in `cache'
	34: from /Library/Ruby/Gems/2.6.0/gems/licensed-2.10.0/lib/licensed/cli.rb:83:in `run'
	33: from /Library/Ruby/Gems/2.6.0/gems/licensed-2.10.0/lib/licensed/commands/cache.rb:23:in `run'
	32: from /Library/Ruby/Gems/2.6.0/gems/licensed-2.10.0/lib/licensed/commands/command.rb:22:in `run'
	31: from /Library/Ruby/Gems/2.6.0/gems/licensed-2.10.0/lib/licensed/reporters/reporter.rb:64:in `report_run'
	30: from /Library/Ruby/Gems/2.6.0/gems/licensed-2.10.0/lib/licensed/commands/command.rb:27:in `block in run'
	29: from /Library/Ruby/Gems/2.6.0/gems/licensed-2.10.0/lib/licensed/commands/command.rb:27:in `map'
	28: from /Library/Ruby/Gems/2.6.0/gems/licensed-2.10.0/lib/licensed/commands/command.rb:27:in `block (2 levels) in run'
[...]

PR #280, opened 2020-06-01
Ensure that nuget dependency path exists before reading

jonabc:
A corollary to https://github.com/github/licensed/pull/273 to check that the path to a nuget dependency exists before calling `File.read`.  This is separated from the referenced PR as that is waiting on some [feedback](https://github.com/dotnet/sdk/issues/11773) from the dotnet team and I'd like to put out a new release soon.

Besides making the ☝️ aforementioned fix, this also slightly changes the functionality of the dependency to be just a bit more performant when:
1. not using the `cache` command
2. using the `cache` command when no changes are needed to cached metadata

These make up the broad majority of `licensed` runs, so hopefully this makes things a little bit faster.  Two specific changes here:
1. Don't get the `project_url` or `description` until evaluating a dependency's `license_metadata`.  That function is only called when caching new metadata or overwriting existing metadata, and takes reading and parsing `nuspect_path` out of the common path.
2. Move checks on `nuspec_path` and `nuspec_contents` into each cached `begin/end` block. This has the effect of caching a nil value rather than not caching a value + re-porforming those checks when they evaluate to false.

/cc @paveliak @zarenner FYI on parceling out some of the fix from the referenced PR

### DIFF
--- a/lib/licensed/sources/nuget.rb
+++ b/lib/licensed/sources/nuget.rb
@@ -17,10 +17,13 @@ module Licensed
         PROJECT_URL_REGEX = /<projectUrl>\s*(.*)\s*<\/projectUrl>/ix.freeze
         PROJECT_DESC_REGEX = /<description>\s*(.*)\s*<\/description>/ix.freeze
 
-        def initialize(name:, version:, path:, search_root: nil, metadata: {}, errors: [])
-          super(name: name, version: version, path: path, search_root: search_root, metadata: metadata, errors: errors)
-          @metadata["homepage"] = project_url if project_url
-          @metadata["summary"] = description if description
+        # Returns the metadata that represents this dependency.  This metadata
+        # is written to YAML in the dependencys cached text file
+        def license_metadata
+          super.tap do |record_metadata|
+            record_metadata["homepage"] = project_url if project_url
+            record_metadata["summary"] = description if description
+          end
         end
 
         def nuspec_path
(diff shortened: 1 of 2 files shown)


lib/licensed/sources/nuget.rb
```diff
@@ -17,10 +17,13 @@ class NuGetDependency < Licensed::Dependency
...
-          super(name: name, version: version, path: path, search_root: search_root, metadata: metadata, errors: errors)
-          @metadata["homepage"] = project_url if project_url
-          @metadata["summary"] = description if description
+        # Returns the metadata that represents this dependency.  This metadata
+        # is written to YAML in the dependencys cached text file
+        def license_metadata
+          super.tap do |record_metadata|
+            record_metadata["homepage"] = project_url if project_url
+            record_metadata["summary"] = description if description
+          end
```
zarenner, line 26: Nice! I wanted to do something like this but I think wasn't sure if it was cool to override `license_metadata`. Also TIL about `tap`, thanks for the knowledge 👍

paveliak: Thinking of it in terms on C# is it akin to overriding virtual property? That is:
1) Get property value from base class
2) Assign two metadata values
3) Return modified property

jonabc: @paveliak yup 👍 .  In ruby all methods can be overridden by default, though there is no special syntax like `virtual` and `override` which explicitly tell which methods can be overridden and which methods are overriding base classes.  The call to `super` here is equivalent to calling `base.licensed_metadata()` in c#.

Issue #286, opened 2020-06-17
Add CLI arg to override which sources to use in a particular scenario


jonabc:
This could be useful as a performance bump when running licensed manually when:
1. a project references dependencies from multiple source types (bundler, npm, etc)
2. automation ties a new install to running licensed, e.g. `bundle install` is hooked to call `licensed cache`

In the above scenario there's no reason to recache or reevaluate licenses for anything but bundler, and a CLI option seems like a minimal lift to help out here.

The only thing this needs to really be aware of is that existing licenses will get deleted for sources that are not evaluated.  Any solution needs to bypass cached data from other sources 👍

sergey-alekseev:
Great idea! I released https://rubygems.org/gems/bundler-licensed/versions/0.2.0 which uses this feature.

PR #287, opened 2020-06-19
Add "sources" CLI arg to filter which sources are run during a command

jonabc:
closes https://github.com/github/licensed/issues/286

This PR adds an array `-s/--sources` option to the `cache`, `status`, `list` and `notices` commands to restrict which sources those commands operate over.

The purpose of this is to accommodate repos that use dependencies from multiple sources but might only update one source at a time.  For example you could run `bundle install && licensed cache --source bundler`, or set up a bundler post hook to run caching anytime packages are installed.

This also fixes a subtle bug that was introduced when enabling shared cache paths.  In prior versions, it was safe to put extra data under an applications `cache_path`, as long as it wasn't in an evaluated sources cache subdirectory.  With that change any files under `cache_path` that did not match to evaluated dependencies would be deleted.  With this PR, that behavior is reverted 👍

### DIFF
--- a/lib/licensed/cli.rb
+++ b/lib/licensed/cli.rb
@@ -30,8 +30,11 @@ module Licensed
     desc "list", "List dependencies"
     method_option :config, aliases: "-c", type: :string,
       desc: "Path to licensed configuration file"
+    method_option :sources, aliases: "-s", type: :array,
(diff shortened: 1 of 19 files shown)


jonabc:
Tested against a local repo that I wanted to use this with, looks good 👍 !

PR #292, opened 2020-07-06
Eliminate ruby 2.7 keyword arg warnings in cli

jurre:
Ruby 2.7 gives deprecation warnings for keyword arguments being passed as a hash.
This change removes a few of those warnings.

### DIFF
--- a/lib/licensed/cli.rb
+++ b/lib/licensed/cli.rb
@@ -14,7 +14,7 @@ module Licensed
       desc: "Individual source(s) to evaluate.  Must also be enabled via configuration."
     def cache
       run Licensed::Commands::Cache.new(config: config),
-          { force: options[:force], sources: options[:sources] }
+          force: options[:force], sources: options[:sources]
     end
 
     desc "status", "Check status of dependencies' cached licenses"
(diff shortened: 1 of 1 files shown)


jurre:
I don't think so, there are a few other warnings:

```
test/cli_test.rb:28: warning: assigned but unused variable - status
test/cli_test.rb:52: warning: assigned but unused variable - status
test/cli_test.rb:76: warning: assigned but unused variable - status
test/cli_test.rb:87: warning: assigned but unused variable - status
```

And a few warnings in gems:

```
licensee-9.14.0/lib/licensee/hash_helper.rb:16: warning: mismatched indentations at 'end' with 'else' at 14
licensee-9.14.0/lib/licensee/matchers/dice.rb:13: warning: mismatched indentations at 'end' with 'else' at 11
licensee-9.14.0/lib/licensee/license.rb:121: warning: method redefined; discarding old spdx_id
lib/ruby/2.7.0/forwardable.rb:226: warning: previous definition of spdx_id was here
```

I did have an issue locally `WARNING: Nokogiri was built against LibXML version 2.9.10, but has dynamically loaded 2.9.4` that I didn't bother looking into 😬 so I might have missed something.

jonabc:
> I don't think so, there are a few other warnings:

I've been fixing these as I find them, thanks for the heads up I'll fix those.

> And a few warnings in gems:

👍  if/when those are fixed licensed can update references

> I did have an issue locally WARNING: Nokogiri was built against LibXML version 2.9.10, but has dynamically loaded 2.9.4 that I didn't bother looking into 😬 so I might have missed something.

I think that error is dependent on the version of libxml you have installed locally and isn't directly related to the codebase 👍

jurre:
> I think that error is dependent on the version of libxml you have installed locally and isn't directly related to the codebase 

Yeah that's just my local dev env 👍

Issue #304, opened 2020-09-23
Licensed crashing when saved YAML files can't be parsed


jonabc:
Noticed this happen a few times where license files were edited manually and ended up having YAML text block specifiers like  `+` or `2` characters removed from `|+` or `|2` respectively.

Failure to parse a yaml file needs to be handled gracefully.

PR #306, opened 2020-09-23
Handle yaml parsing errors when reading dependency records from disk

jonabc:
Closes https://github.com/github/licensed/issues/304

Adding a rescue for `Psych::SyntaxError` when reading DependencyRecords from disk, which raises the error message to a new internal error type `DependencyRecord::Error`.  That internal error is caught and handled in the command class, with the error message output similar to an error encountered while shelling out a command.


jonabc:
core failures are being handled in a separate branch

Issue #243, opened 2020-02-28
Yarn source fails if dependency was installed in a devDependency location


gareth:
I'm not *entirely* sure what's going on here, so I'll describe what I've seen and what I'm guessing is happening.

The `yarn` source isn't always able to find its dependency locations. It makes an assumption that dependencies are always found inside their parent dependency:

https://github.com/github/licensed/blob/ef02cc78cd4b24b31d18596be0530b086b9879c0/lib/licensed/sources/yarn.rb#L65

However, this doesn't seem to be true. If Yarn has already installed an identical version of a dependency, it seems to put later copies directly inside the top-level `node_modules` folder (see the end of my reproduction steps below). So the correctly nested node_module only seems to exist inside one dependency.

If the same version of a nested dependency is required by both a production dependency and a development dependency, it's possible that the node_module is installed only inside the development dependency.

This causes problems because the `yarn` source adds `--production` to its list command, which means it's impossible for it to find a nested dependency that doesn't exist inside the installed production dependencies.

Note this problem isn't solved by setting the `production_only` option to false: I **do** only want to generate license conflicts for _production dependencies_, it's just that yarn may only have stored that license information inside a _development dependency's_ `node_modules`.

It was tricky coming up with a minimal test case for this, but this seems to demonstrate it. It also shows that the `npm` source doesn't seem to have this problem (it's correctly able to find the right license info):

```
$ yarn --version
1.22.0
$ licensed --version
2.8.0
$ mkdir licensed-test; cd licensed-test
$ printf "sources:\n  npm: true\n  yarn: true" > .licensed.yml
$ yarn add -s --no-progress babel-loader@8.0.6
warning " > babel-loader@8.0.6" has unmet peer dependency "@babel/core@^7.0.0".
warning " > babel-loader@8.0.6" has unmet peer dependency "webpack@>=2".
$ yarn add -D -s --no-progress vue-jest@3.0.5
warning package.json: No license field
warning No license field
warning vue-jest > babel-plugin-transform-es2015-modules-commonjs > babel-runtime > core-js@2.6.11: core-js@<3 is no longer maintained and not recommended for usage due to the number of issues. Please, upgrade your dependencies to the actual version of core-js@3.
warning " > babel-loader@8.0.6" has unmet peer dependency "@babel/core@^7.0.0".
warning " > babel-loader@8.0.6" has unmet peer dependency "webpack@>=2".
[...]

Issue #324, opened 2020-11-01
license type "none" results in "missing license text" error


markphelps:
See: https://github.com/markphelps/flipt/runs/1338656481
Code: https://github.com/markphelps/flipt/compare/ui-licenses

## Problem

Example License:

```
---
name: ansi-regex
version: 4.1.0
type: yarn
homepage: https://github.com/chalk/ansi-regex.git
license: none
licenses: []
notices: []
```

Running `licensed status`

Error:

```
Errors:
* ui.yarn.ansi-regex-4.1.0
  filename: /Users/markphelps/workspace/flipt/.licenses/ui/yarn/ansi-regex-4.1.0.dep.yml
    - missing license text
```

Also ran `licensed cache --force && licensed status` to no avail

## Expected Behavior

Licenses with type `none` would not return an error `missing license text` when running `licensed status`

jonabc:
@markphelps I'm not sure this is a safe change to make as it removes a critical check in ensuring compliance.  In your branch you've set `none` as an allowed license type, I'm assuming to get CI to pass.  That solves the immediate failures but defeats the purpose of ensuring that all dependencies use allowed OSS licenses.

Licensed depends on all dependency packages being available locally in order to detect license contents.  In order for it to work properly in CI, you should add in steps before running licensed-ci in your licensed CI workflow
1. setup node
2. install yarn globally
3. run yarn to install dependencies for the project

I'm going to close this, but I'm happy to help you get your CI workflow passing.  Let me know how I can help.

markphelps:
@jonabc , yes that would be great if you wouldnt mind helping

I tried to follow your suggestions locally to see if yarn would pull the dependencies locally, but it didnt seem to change anything:

```
$ npm install -g yarn --force
$ cd ui
$ yarn
yarn install v1.22.10
[1/6] 🔍  Validating package.json...
[2/6] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.54s.
$ git diff
[ no output ]
```

jonabc:
@markphelps sorry for the delay.  I cloned your repo locally to take a look at the ui-licenses branch and I think you're running into https://github.com/github/licensed/issues/243.  I tried the [workaround](https://github.com/github/licensed/issues/243#issuecomment-594592514) in that issue, and licensed was able to find all dependency locations on disk using the npm source.  Note that this doesn't force you to use npm to manage your dependencies - try out the config below.

[...]

PR #326, opened 2020-11-04
Find yarn dependency package paths via brute force

jonabc:
Closes https://github.com/github/licensed/issues/243

After the linked bug had another hit as the underlying issue in https://github.com/github/licensed/issues/324, I took another look and couldn't find any logical way that made sense to detect yarn package paths under node_modules based on the output from any yarn CLI.

As an alternative, here's a solution that finds and parses every package.json file under node_modules to make a map of all "name@version" => directory pairs.  The map is cached so it's at least only evaluated once.

The performance impact isn't too noticeable for small to mid-size projects, but it gets a little more expensive for larger projects.  Still, it seems to only take a few seconds for larger projects so it's not too bad.

### DIFF
--- a/lib/licensed/sources/yarn.rb
+++ b/lib/licensed/sources/yarn.rb
@@ -36,7 +36,7 @@ module Licensed
       def packages
         return [] if yarn_package_tree.nil?
         all_dependencies = {}
-        recursive_dependencies(config.pwd, yarn_package_tree).each do |name, results|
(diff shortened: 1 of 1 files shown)


jonabc:
Moving forward with this solution for now - the yarn source won't work well without this sort of change and the performance hit isn't too bad except for projects with a very large number of dependencies.

PR #328, opened 2020-12-09
`generated_license_contents`: early return when `license.text` is nil

Eun:
I have this error when running `licensed` (in a larger project):
```
/licensed-2.14.2/lib/licensed/dependency.rb:149:in `generated_license_contents': undefined method `lines' for nil:NilClass (NoMethodError)
```

I figured that `license.text` in [dependency.rb:149](https://github.com/github/licensed/blob/2.14.2/lib/licensed/dependency.rb#L149) must be `nil` for this error to occur. 

I am not a ruby developer so feel free to edit the code as much as needed.

### DIFF
--- a/lib/licensed/dependency.rb
+++ b/lib/licensed/dependency.rb
@@ -142,6 +142,7 @@ module Licensed
     def generated_license_contents
       return unless license
       return if license.key == "other"
+      return unless license.text.nil?
 
       # strip copyright clauses and any extra newlines
       # many package managers don't provide enough information to
--- a/lib/licensed/dependency.rb
+++ b/lib/licensed/dependency.rb
@@ -142,7 +142,7 @@ module Licensed
     def generated_license_contents
       return unless license
       return if license.key == "other"
-      return unless license.text.nil?
+      return if license.text.nil?
 
       # strip copyright clauses and any extra newlines
       # many package managers don't provide enough information to


lib/licensed/dependency.rb
```diff
@@ -142,6 +142,7 @@ def license_metadata
     def generated_license_contents
       return unless license
       return if license.key == "other"
+      return unless license.text.nil?
```
jonabc, line 145: I think this is backwards, the function should return when the text is nil.

```suggestion
      return if license.text.nil?
```

jonabc (commented):
Interesting, thanks for catching this and opening the PR!  Do you have a reproduction scenario or project you can share, or do you know which dependency this occurs on?

It would be interesting to know when this error occurs because it's strange to have a detected license, but that license has no text.

Eun:
@jonabc I cannot share the project. But I see what I can do to narrow down the problem.

jonabc:
> I cannot share the project. But I see what I can do to narrow down the problem.

Can you set up a repro project that uses the same dependencies but doesn't have any private code/IP?

Eun:
Was able to reproduce in a controlled environment.
Here are the needed files:

https://gist.github.com/Eun/8afcca0b6ad60ab6fee20d8fbe1f4d1b

```
$ yarn --version
1.22.5
$ node --version
v12.20.0
```

To reproduce do 
```
$ yarn install
$ licensed cache
```

jonabc:
@Eun sorry for the delay.  I was unable to reproduce the issue using the following app versions.

```
➜  repro-licensed-328 git:(master) ✗ yarn -v
1.22.4
➜  repro-licensed-328 git:(master) ✗ npm -v
6.14.4
➜  repro-licensed-328 git:(master) ✗ node -v
v14.1.0
➜  repro-licensed-328 git:(master) ✗ licensed -v
2.14.2
```

I don't want to continue to block you from using the tool, so I'm going to merge this change and release a new version with the fix.  I'd like to continue to understand how you hit this scenario though.  If you wouldn't mind, can you share what versions of yarn, npm, node and licensed you're using locally that are hitting the issue?

jonabc (approved):

Eun:
I stitched a `Dockerfile` together that can be used to reproduce the bug:
```
# This Dockerfile should demonstrate the isse at https://github.com/github/licensed/pull/328
# To build run:
# 	docker build --no-cache -t test:latest
# After building start docker container using
# 	docker run -ti -v $(pwd):/app/ -w /app --rm test:latest /bin/bash
# Inside the docker container run
#   yarn install && licensed cache
FROM ubuntu:20.10

RUN apt-get update && \
	apt-get install -y curl gnupg && \
	curl -sL https://deb.nodesource.com/setup_12.x | bash - && \
	curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
	echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
	curl -sSL -o /tmp/licensed.tar.gz https://github.com/github/licensed/releases/download/2.14.2/licensed-2.14.2-linux-x64.tar.gz && \
	tar -xf /tmp/licensed.tar.gz -C /tmp && \
	mv /tmp/licensed /bin/ && \
	chmod 0777 -R /tmp && \
	apt-get update && \
	apt-get install -y nodejs yarn

```

jonabc:
@Eun thanks for the dockerfile, I was able to figure out why I was not reproducing the issue locally and ultimately get the root cause.  The `time-input` dependency is listed as `UNLICENSED` in it's package.json, which licensee translates to the `no-license` license that has no text.  Checking for nil text is the right fix 👍 

Thanks for the opening the PR and working with me to validate the root cause 🙇

Issue #333, opened 2021-02-03
Having trouble with setting up licensed "missing license text"


lowlighter:
Hello!

I followed the docs but I'm still having some trouble figuring how to make it work 😕

Here's my `.licensed.yml` config file:
```yaml
source_path: .
cache_path: .licensed
```

And below the commands I ran (basically trying to run **licensed** over a NodeJS repository):
```shell
git clone https://github.com/user/test # example 
cd test
npm ci
vi .licensed.yml 
licensed cache
licensed status
```

`licensed cache` seems to works but I'm getting the following errors on `licensed status`:
```diff
- missing license text
- license needs review: none

209 dependencies, 418 errors found # i.e. all dependencies
```

#324 seemed to have a similar error (although from what I understood it was for the GitHub action version, I'm trying to run the executable), so I added `allowed: [none]` and it make all `license needs review: none` errors disappears, but I'm still encountering *missing license text*.

Seems licenses aren't correclty detected as they exists:
```
ls node_modules/axios
CHANGELOG.md dist index.d.ts index.js lib LICENSE package.json README.md UPGRADE_GUIDE.md
```
But cached files are empty (all of them): 
```yaml
---
name: axios
version: 0.21.1
type: npm
summary:
homepage:
license: none
licenses: []
notices: []
```

What I am missing?

___

Also, is there any example output? I'm not sure yet that **licensed** is the tool that I need for what I want to achieve, I may have misunderstood its utility.

Thanks a lot!

jonabc:
@lowlighter it looks like licensed is unable to find any of the dependencies on disk even after you've run npm ci.  In that same environment after running `npm ci` can you run `npm list --json --long --production` and check:
1. does the command succeed?  if there are errors, what are the errors?
2. do returned packages' `path` property point to a valid location on disk that contains the the installed dependency?

> Also, is there any example output? I'm not sure yet that licensed is the tool that I need for what I want to achieve, I may have misunderstood its utility.

The output that you've seen is the 90% case of what licensed does, with the exception that the files which are currently empty should not be empty 😅 .  If you can share what you're trying to achieve I may be able to tell you whether licensed is a good fit or not.  Please let me know if you have any other questions as well 🙇

lowlighter:
Running `npm list --json --long --production` was displaying some warnings before JSON output (npm/NodeJS from apt repo were outdated on my vm). Installing them from nodesource to be up-to-date solved the problem 👍

[...]

PR #334, opened 2021-02-08
Add more output options for list and cache commands

jonabc:
This branch makes a few relatively small changes as followup to some quick exploration while looking at an [issue report](https://github.com/github/licensed/issues/333#issuecomment-772891527)

> This got me curious and I've put together a quick change locally that could give something like the above output by (1) extending licensed list to support different output formats and (2) including more information in the output. That unfortunately has some performance implications to that command so I'd need to review and see how to make the new functionality available without degrading general usage.

1. Adds the "format" CLI flag to the list and cache commands and passes the flag into each commands' `run` call.  Usage of the flag is already handled by the base command class so no further changes are needed to support JSON and YML output
2. Sets some dependency metadata into the `list` commands report object to be included in JSON/YML output
3. Adds a new CLI flag `-l/--licenses` to the list command that includes the detected license key in the output.
   - the default list output will look like `name (version): license`
   - included in json and YML as the `license` key on each dependency
   
/cc @lowlighter FYI

### DIFF
--- a/lib/licensed/cli.rb
+++ b/lib/licensed/cli.rb
@@ -12,9 +12,11 @@ module Licensed
       desc: "Path to licensed configuration file"
     method_option :sources, aliases: "-s", type: :array,
       desc: "Individual source(s) to evaluate.  Must also be enabled via configuration."
+    method_option :format, aliases: "-f", enum: ["yaml", "json"],
+      desc: "Output format"
     def cache
       run Licensed::Commands::Cache.new(config: config),
-          force: options[:force], sources: options[:sources]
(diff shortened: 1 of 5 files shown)


lowlighter:
Awesome!

Is there an ETA yet for v2.14.4 ?

jonabc:
There's not but there's nothing to stop cutting a new release today/tomorrow if it would be useful for you

lowlighter:
That would be nice, though I'm not in a hurry so if there are still some stuff to implement toward 2.14.4 I can wait 🙂

PR #335, opened 2021-02-09
Release 2.14.4

jonabc:
## 2.14.4
2021-02-09

## Added
- `list` and `cache` commands optionally print output in JSON or YML formats using the `--format/-f` flag (https://github.com/github/licensed/pull/334)
- `list` command will include detected license keys using the `--licenses/-l` flag (https://github.com/github/licensed/pull/334)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## 2.14.4
+2021-02-09
+
+## Added
+- `list` and `cache` commands optionally print output in JSON or YML formats using the `--format/-f` flag (https://github.com/github/licensed/pull/334)
+- `list` command will include detected license keys using the `--licenses/-l` flag (https://github.com/github/licensed/pull/334)
+
 ## 2.14.3
 2020-12-11
 
@@ -379,4 +386,4 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 Initial release :tada:
 
-[Unreleased]: https://github.com/github/licensed/compare/2.14.3...HEAD
+[Unreleased]: https://github.com/github/licensed/compare/2.14.4...HEAD
--- a/lib/licensed/version.rb
+++ b/lib/licensed/version.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 module Licensed
-  VERSION = "2.14.3".freeze
+  VERSION = "2.14.4".freeze
 
   def self.previous_major_versions
     major_version = Gem::Version.new(Licensed::VERSION).segments.first


jonabc:
Ignoring the actions failures - there is no code changing in this project and I think the majority of failures are probably related to https://github.com/actions/virtual-environments/issues/1816

jonabc:
Hah actually, following on ☝️ it'll be interesting to see if the release action works...

lowlighter:
There are no executable yet linked to current release (seems that [mac build](https://github.com/github/licensed/actions/runs/553335420) failed on a ssl certificate error which prevented packages upload) and [`gem install licensed`](https://rubygems.org/gems/licensed/versions/2.14.3) is still limited to 2.14.3 🙂

jonabc:
@lowlighter yep I was playing around with the build failure last night but wasn't able to get it working.  If you have ideas on how to fix the SSL errors on MacOS, I'd appreciate the extra 👀 😄

lowlighter:
```
ERROR:  You must add /OU=GlobalSign Root CA - R3/O=GlobalSign/CN=GlobalSign to your local trusted store
ERROR:  Could not find a valid gem 'reverse_markdown' (~> 1.0), here is why:
          Unable to download data from https://rubygems.org/ - SSL_connect returned=1 errno=0 state=error: certificate verify
          failed (unable to get local issuer certificate) (https://rubygems.org/specs.4.8.gz)
```
From [these logs](https://github.com/github/licensed/runs/1868244868?check_suite_focus=true#step:4:2652) I would be tempted to say that it's missing the certificate authority which authenticate `rubygems.org`.

I'm not an expert with Ruby and MacOS, but this [rubygems issue #1758](https://github.com/rubygems/rubygems/issues/1758) might help (though it seems a bit old).
Seems that lucky ones were able to fix it with `rvm osx-ssl-certs update all`. There also might be some errors with [/etc/openssl/cert.pem and /usr/local/etc/openssl/cert.pem](https://github.com/rubygems/rubygems/issues/1758#issuecomment-262478846) (basically OpenSSL path/version caused weird behaviour).

I'm afraid I won't be that much help on this 😅
Good luck

jonabc:
> I'm afraid I won't be that much help on this 😅
> Good luck

😆 😢  haha thanks

jonabc:
As a quick update here, I thought I had gotten builds working in https://github.com/github/licensed/pull/336, but the binaries built using that branch (at least the linux package) are [not working](https://github.com/github/licensed/issues/337).  I've pulled those binaries from the release and will look into fixing the build process again

lowlighter:
Good to know 🙂

I've only used the latest published [gem](https://rubygems.org/gems/licensed/versions/2.14.4) and did not encounter any issues  (WSL Debian 10) but I've just played with `licensed list --format=json --licenses` (which works really great by the way) until now. 

Not sure if executable not working implies that gem may have issues (it may worth a check too)

jonabc:
The gem should be fine.  The error is with the way the executables are built and not necessarily with the gem.  I'd expect that if in the exact same linux environment someone installed ruby and ran the gem everything would work

PR #336, opened 2021-02-10
Fix the release scripting for mac builds

jonabc:
This fixes the package build failures that were [called out](https://github.com/github/licensed/pull/335#issuecomment-776816271) on the 2.14.4 release issue.

A few changes were needed to fix SSL failures during the mac build process.  The SSL failures were not present on the build host, but were coming up during the build process.  The project uses https://github.com/kontena/ruby-packer to build the ruby code into a redistributable package, but the inner ruby build thats included in the package doesn't include certs by default as mentioned in the [README](https://github.com/kontena/ruby-packer#openssl).

I've done a few things here to fix the situation
1. Updated the release workflow to use `ruby/setup-ruby` instead of `actions/setup-ruby`, to make sure that the most recent and up to date ruby code was being brought onto the build host
1. Included `gem update --system` and `gem update bundler` in the mac build script, again to make sure that the most recent versions of each are being used in order to update any certificates or expectations related to certificates
1. Set the `--openssl-dir` argument when running rubyc to the parent directory of ruby's openssl default certificate directory.

With all of the above, I was able to successfully build both the linux and mac distributable exes which I minimally tested to ensure they were working.  I've attached those built packages to the [2.14.4 release](https://github.com/github/licensed/releases/tag/2.14.4) and released the built gem to rubygems

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -93,9 +93,9 @@ jobs:
         fetch-depth: 0
 
     - name: Set up Ruby 2.6
-      uses: actions/setup-ruby@v1
+      uses: ruby/setup-ruby@v1
       with:
-        ruby-version: 2.6.x
+        ruby-version: 2.6
 
     - name: Build package
       run: script/packages/linux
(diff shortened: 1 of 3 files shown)


jonabc:
CI failures are known and unrelated to this change.  Going to move ahead with merging this to unblock future releases.

Issue #337, opened 2021-02-11
Licensed fails with XCRYPT_2.0 and GLIBC_2.29 not found


johlju:
This error started today, yesterday `licensed` ran without issue. What dependency is missing on the build worker? 

Running a GitHub Action job like this.

```yaml
jobs:
  License:
    name: Valid licenses
    runs-on: ubuntu-latest
    steps:
    - name: Checkout
      uses: actions/checkout@v2
      with:
        fetch-depth: 0
    - name: Setup Licensed
      uses: jonabc/setup-licensed@v1
      with:
        version: 2.x
    - name: Install build dependencies
      run: |
        sudo npm install --global npm
        npm install
    - name: Licensed
      id: licensed
      run: |
        licensed version
```
Ends up with the following when running step id `licensed`:

```plaintext
licensed: /lib/x86_64-linux-gnu/libcrypt.so.1: version `XCRYPT_2.0' not found (required by licensed)
licensed: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by licensed)
```

I tried adding the following but it did not help:

```
    - name: Install dependencies
      run: |
        sudo apt-get install cmake pkg-config
```

GitHub Actions job have this information for the build worker:

```
Operating System
  Ubuntu
  18.04.5
  LTS
Virtual Environment
  Environment: ubuntu-18.04
  Version: 20210131.1
  Included Software: https://github.com/actions/virtual-environments/blob/ubuntu18/20210131.1/images/linux/Ubuntu1804-README.md
```

Thank you for a great tool!

jonabc:
@johlju oof ok I'll take a look.

I released a new version of insights yesterday which included minor functionality updates to output on the list and cache commands - nothing major.  I tested those builds on ubuntu but didn't encounter this 😢 

I'm going to pull the 2.14.4 release binaries for the time being since I'm guessing this is going to affect many users.

> Thank you for a great tool!

❤️

johlju:
Ran the build again and it now used `2.14.3`  which ran without any issues. I saw that it was probably due to a problem in the build pipeline when the new version was release. Thanks for the quick workaround.

jonabc:
[...]

PR #338, opened 2021-02-12
Force linux packaging script onto 18.04

jonabc:
closes https://github.com/github/licensed/issues/337

It looks like the underlying issue is that the the version of ruby-packer I'm using to build the licensed executable packages may not be compatible with ubuntu-20.04.  I think this repo might have been affected early for the change of the `ubuntu-latest` host os in actions to run on 20 instead of 18.  I can't find any documentation about the general switch for ubuntu-latest to 20.04 so I'm not sure if it's happened for all repos.  This repo might have been affected sooner to beta test the changes due to being a GitHub property 🤷 

The package build is not the only thing affected either, many of the CI tests using actions are broken as well due to not finding packages or frameworks available no the new machine type.

I forced the linux packaging job back to using ubuntu-18.04 and tested the [built linux package](https://github.com/github/licensed/actions/runs/559467175).  Everything appeared to be working across multiple OS flavors, which was also [validated](https://github.com/github/licensed/issues/337#issuecomment-778139954) by @johlju who opened the linked issue.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -83,7 +83,7 @@ jobs:
 
   package_linux:
     needs: vars
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v2
       with:


jonabc:
I'm expecting many of the CI jobs to fail on this, and will not be waiting for 💚  before merging

Issue #340, opened 2021-02-13
Fix npm source for npm 7+


jonabc:
Looks like `npm list` changed functionality with npm 7 so that `npm list` will only [print top-level dependencies by default](https://docs.npmjs.com/cli/v7/commands/npm-ls#note-design-changes-pending).

There is a new `--all` flag which can be used to keep the necessary functionality for this tool, but should only be used with npm 7+

PR #341, opened 2021-02-13
Support npm 7 in npm source

jonabc:
closes https://github.com/github/licensed/issues/340

With NPM 7, the `npm list` command is changed to no longer return the full dependency tree unless `--all` is specified.  This branch/PR checks whether to add the argument to the list command based on the output from `npm -v`, parsed and compared using `Gem::Version`.

I'm updating the CI workflows first to run npm CI tests on node 14 and 15, where 15 includes npm 7 by default.  I'm expecting that without the changes, we'd see tests on indirect dependencies to fail with node 15 / npm 7.


jonabc:
😞  this turned into a much larger change than I expected it to.  Along with needing the `--all` flag, there are a few other changes I found that were needed
1. `npm list --long` no longer includes homepage and description in the returned data 😢 .  Why?  Who knows, probably performance considerations.  To retain that information I need to load and parse each dependency's package.json file
2. Missing dependencies raise an error where they used to just be marked as missing.  Accounting for this by enhancing the error message when the result from `npm list ...` can't be parsed to include the returned error message.  The implementation isn't amazing but it avoids needing a refactoring or extension of the shell execution code.

jonabc:
Many of the CI checks are still failing, but node15/npm7 is successful.  Going to merge this though I'm still not in love with all the checks for how to handle different npm versions.  This might see some iteration in the future.

PR #343, opened 2021-03-04
Fix spelling of npm

lumaxis:
See https://github.com/npm/cli#is-it-npm-or-npm-or-npm 🙂 

I thought about also renaming the class but lowercase classnames are also odd in Ruby, so being pragmatic there 😄


jonabc (approved):
👍  thanks!

isaacs:
It is acceptable to spell npm as "NPM" in cases which are traditionally all-caps or need to be capitalized for some other reason.  I think Ruby class definitions qualify ;)

PR #350, opened 2021-04-06
pip: fix regex: match packages with dots in name

bcskda:
Ran into this in a python-2.7 pipeline:

https://github.com/bcskda/anytask/runs/2280835856


jonabc (approved):
❤️  looks great, thank you!

Issue #353, opened 2021-04-09
mini_portile gem and its LICENSE.txt not being detected


pjmartorell:
Hi, when installing `mini_portile2` (2.5.0) and running `licensed cache -s bundler` is not detecting the license of the gem even it has it (https://github.com/flavorjones/mini_portile).

jonabc:
:wave: @pjmartorell I wasn't able to reproduce this with the following:
1. add `gem "mini_portile2", "2.5.0"` to [test/fixtures/bundler/Gemfile](https://github.com/github/licensed/blob/master/test/fixtures/bundler/Gemfile)
2. run `script/source-setup/bundler`
3. run `licensed list -c test/fixtures/command/bundler.yml` <- verify that mini_portile2 is listed
4. run `licensed cache -c test/fixtures/command/bundler.yml` <- verify that mini_portile2 data is cached at `test/fixtures/bundler/.licenses/bundler/mini_portile2.dep.yml`

If you're still having the problem can you verify that the gem is installed locally for the current bundler scope or gem installation path?  If you've verified the issue is persisting can you set up a small project repo or Gemfile that reproduces the problem?

Some other information that might be helpful are what versions of ruby and/or bundler you're using when you encounter the problem.

pjmartorell:
Hi @jonabc! thanks for the fast reply, I attach a [repo](https://github.com/pjmartorell/test) with a small project where I just:
- created a rails app with `rails new _6.0.3.5_ testapp` with ruby 2.7 
- added `licensed` gem and `.licensed.yml` config
- run `licensed cache -s bundler` and checked `mini_portile2` is not present even it's a dependency of a gem of the default/production group in the Gemfile.

- Ruby version 2.7, but I think it is not meaningful since it was happening the same with ruby 2.5 and 2.6
- Bundler 2.1.4

jonabc:
@pjmartorell thanks for the report and test repo, this is very helpful.  I'm still digging in, but the code for the bundler source is tricky because it's attempting to manipulate bundler, which in turn manipulates rubygems.  It's a bit of a mess that's compounded by trying to make the bundler source usable when run from the non-gem, standalone executable distributions.

[...]

PR #354, opened 2021-04-18
Disable bundler source when running under ruby packer

jonabc:
This PR restricts the bundler source to run only when run as a gem installed on a host machine.  If the bundler source is used from the distributed executables built using ruby-packer, an error will be given to the user.  In the process I've been able to make better use of bundler's high level APIs and significantly reduce the size and complexity of the code in this project.

To support bundler dependency enumeration from the ruby-packer based executable,  I ended up having to write a slimmed-down approximation of bundlers internal logic for enumerating dependencies and matching them up to gem specifications where available.  This has been a source of many bugs and unhandled edge cases, the most recent of which is https://github.com/github/licensed/issues/353.

Keeping the bundler source working has been challenging to say the least due to needing the source to work from both a gem installation in a ruby environment and from within the ruby_packer environment of the distributable licensed exe files that are built and attached to releases.

Enabling bundler dependency enumeration from the distributable licensed exe is mostly for convenience.  If someone is enumerating bundler dependencies then it's not a stretch to assume they have a working ruby installation.  This tool requires that dependencies have been installed locally, and it seems like a far edge case for anyone to install gems on a system then uninstall ruby 🤷 

The core of the problem with using licensed's bundler source from the distributed exe is that there are effectively two pairs of ruby and bundler that are running at the time - one on the host machine and one inside the distributed executable.  Bundler was not built to support the use case of having two concurrently running ruby and bundler environments because of course it isn't, why would that ever be needed 😆 .  However licensed does need to run in that environment and because it's using bundlers internal objects it needs to trick bundler into doing the same.

Enough is enough, it shouldn't be too much to ask to run `gem install licensed` if a user is going to be enumerating bundler dependencies 😆

### DIFF
--- a/docs/sources/bundler.md
+++ b/docs/sources/bundler.md
@@ -2,17 +2,7 @@
 
 The bundler source will detect dependencies `Gemfile` and `Gemfile.lock` files are found at an apps `source_path`.  The source uses the `Bundler` API to enumerate dependencies from `Gemfile` and `Gemfile.lock`.
 
-### Enumerating bundler dependencies when using the licensed executable
-
-**Note** this content only applies to running licensed from an executable.  It does not apply when using licensed as a gem.
-
-_It is required that the ruby runtime is available when running the licensed executable._
-
-The licensed executable contains and runs a version of ruby.  When using the Bundler APIs, a mismatch between the version of ruby built into the licensed executable and the version of licensed used during `bundle install` can occur.  This mismatch can lead to licensed raising errors due to not finding dependencies.
-
-For example, if `bundle install` was run with ruby 2.5.0 then the bundler specification path would be `<bundle path>/ruby/2.5.0/specifications`.  However, if the licensed executable contains ruby 2.4.0, then licensed will be looking for specifications at `<bundle path>/ruby/2.4.0/specifications`.  That path may not exist, or it may contain invalid or stale content.
-
-To prevent confusion, licensed uses the local ruby runtime to determine the ruby version for local gems during `bundle install`.  If bundler is also available, then the ruby command will be run from a `bundle exec` context.
+**Note** The bundler source cannot be used when running the [packaged licensed executable](../packaging.md)
 
 ### Excluding gem groups
 
--- a/lib/licensed/sources/bundler.rb
+++ b/lib/licensed/sources/bundler.rb
@@ -2,56 +2,19 @@
 require "delegate"
 begin
   require "bundler"
+  require "licensed/sources/bundler/missing_specification"
 rescue LoadError
 end
 
 module Licensed
   module Sources
     class Bundler < Source
-      class MissingSpecification < Gem::BasicSpecification
-        attr_reader :name, :requirement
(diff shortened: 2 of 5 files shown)


jonabc:
@pjmartorell if you're able, would you mind installing and using the bundler gem attached to a [pre-release for this change](https://github.com/github/licensed/releases/tag/pre-release-disable-bundler-ruby-packer) in your [test repo](https://github.com/pjmartorell/test) to see if it fixes the issue for you?  It worked for me, but I'd appreciate a second set of 👀 , because this makes a breaking change in order to support better bundler dependency detection.  Thanks!

pjmartorell:
@jonabc I unzipped the zip file in the root of the project and I added the gem to the `Gemfile` with `gem 'licensed', path: './licensed'` and the gem was installed successfully but the `licensed` command is not found.

I also tried to build them and install the `.gem` but it happens the same. Running `bundle exec licensed` or `exe/licensed` didn't work either. Am I doing something wrong?

jonabc:
@pjmartorell that's weird.  if you look at your bundle install path do you see the licensed gem in there, and what is the version in `<path to bundled licensed code>/lib/licensed/version.rb`?

If you're installing licensed from your Gemfile I wouldn't expect `licensed` to be runnable standalone but `bundle exec licensed` should work.

pjmartorell:
@jonabc the version is correct, but executables are not found:
```bash
> gem uninstall licensed

 1. licensed-2.15.2.pre.bundler
 2. licensed-2.15.2
 3. All versions
> 3
Successfully uninstalled licensed-2.15.2.pre.bundler
Remove executables:
	licensed

in addition to the gem? [Yn]  Y
Removing licensed
Successfully uninstalled licensed-2.15.2
> cd licensed
> rake install                                                                                                                                                                                                                                    
licensed 2.15.2.pre.bundler built to pkg/licensed-2.15.2.pre.bundler.gem.
licensed (2.15.2.pre.bundler) installed.
> gem install pkg/licensed-2.15.2.pre.bundler.gem
Successfully installed licensed-2.15.2.pre.bundler
1 gem installed
> rbenv rehash                                                                                                                                                                                                                                    > licensed cache -s bundler                                                                                                                                                                                                                       
zsh: command not found: licensed
> rehash                                                                                                                                                                                                                                     > licensed cache -s bundler
zsh: command not found: licensed
> gem list licensed

*** LOCAL GEMS ***

licensed (2.15.2.pre.bundler)
   
```

jonabc:
@pjmartorell that is really weird and TBH I don't know whats going on.  Two followup things to check:
1. Is the correct rbenv shims folder in your path?
2. Does `rbenv which licensed` succeed and give you a path to licensed?
   - If it does, are you able to run licensed entrypoint via it's absolute path?

jonabc:
I'm going to move forward with this change.  The next release will be a major version update due to the breaking changes in the bundler source here, so hopefully everyone is referencing licensed with some sort of scoping like `~> 2.0` or `2.x`, and upgrading to 3.0.0 will be an intentional choice.

Issue #359, opened 2021-05-15
Error when specifying a bundler gem from a git repo


clinejj:
### Describe the bug

When using a git repo for a bundler gem, licensee reports the following error and any licensed command fails:
```
Search root must be the project path directory or its ancestor
from /Users/john/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/licensed-3.0.0/lib/licensed/dependency.rb:40:in `initialize'
        from /Users/john/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/licensed-3.0.0/lib/licensed/sources/bundler.rb:17:in `initialize'
        from /Users/john/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/licensed-3.0.0/lib/licensed/sources/bundler.rb:60:in `new'
        from /Users/john/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/licensed-3.0.0/lib/licensed/sources/bundler.rb:60:in `block (2 levels) in enumerate_dependencies'
```

My gemfile specifies:
```
gem 'active_admin_import', '>= 4.2.1.ruby3', git: 'https://github.com/Shelf-Life/active_admin_import'
```
At that line, it seems licensed is expecting a different folder than what the gem is installed to:
```
Search root must be the project path directory or its ancestor: /Users/john/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/active_admin_import-4.2.1.ruby3 and /Users/john/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/bundler/gems/active_admin_import-3997e99f222e
```
are defined for root and path at this line: https://github.com/licensee/licensee/blob/master/lib/licensee/projects/fs_project.rb#L23

I had filed with licensee here and was suggested this was an issue with licensed: https://github.com/licensee/licensee/issues/492

This appears to be due to a mismatch between what licensed passes to licensee as the search root for these types of dependencies.

### Steps to reproduce the behavior

1. Add the above gem to a gemfile for a Ruby project
2. Run `bundle install`
3. Run `gem install licensed`
4. Run `licensed cache`

### Expected behavior

`licensed` should run as expected with no errors.

jonabc:
Thanks for the report.  I took a look into the cause and have a small change incoming that should fix this.

PR #360, opened 2021-05-17
Validate git based ruby gems work with bundler source

jonabc:
closes https://github.com/github/licensed/issues/359

The issue here is that the path construction for the search root results in a path that doesn't exist and is not an ancestor for gems pulled via a "git" directive in bundlers gemfile.  The change in this PR ensures that the root path exists before it's returned.  This should be ok for the case given but doesn't fully address the use case for git sourced bundler dependencies and may need to be further addressed in the future.  For now, this is a small change that is good to have regardless.

I validated this change originally using the `gem` line shared in the related issue, but opted to use a smaller gem that pulls in fewer dependencies and is not a fork in the checked in test fixture

cc clinejj

### DIFF
--- a/lib/licensed/sources/bundler.rb
+++ b/lib/licensed/sources/bundler.rb
@@ -12,9 +12,9 @@ module Licensed
       class Dependency < Licensed::Dependency
         attr_reader :loaded_from
 
-        def initialize(name:, version:, path:, loaded_from:, search_root:, errors: [], metadata: {})
+        def initialize(name:, version:, path:, loaded_from:, errors: [], metadata: {})
           @loaded_from = loaded_from
-          super name: name, version: version, path: path, errors: errors, metadata: metadata, search_root: search_root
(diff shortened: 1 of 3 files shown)


jonabc:
I'm not seeing any of these errors in my local environment, digging into that

jonabc:
TL;DR - I'm not sure why I added search_root usage to the bundler source but it appears to not be needed.  Since it was causing issues I removed it.

After doing some debugging both locally and logging values in CI I don't see what value the search_root argument is bringing to bundler dependencies.  Generally when gems get installed they get their own folder where the folder name includes both the name and the version.  Additionally, each gem is packaged and installed as a standalone unit where installed gems don't normally share files with each other.

The search_root argument was added to support the `go` source, where multiple go package imports can come from the same repository and share a LICENSE file at the root of the repo's codebase rather than in each subfolder.  That situation doesn't apply here though and IIRC I used the search_root value to resolve a test failure that no longer seems to be needed?

🤷

Issue #355, opened 2021-04-19
NuGet Source - Support better App Naming when using Glob Patterns


JanMattner:
Currently for the NuGet Source, is is required to set the `source_path` to the object folder of a project, which is typically in `some/path/MyProject/obj` for a project with name `MyProject`. So, a Dotnet project is a single "app" (for GitHub Licensed).

If I have 100 of those, I rather use the glob pattern `some/path/*/obj`.
Problem: all apps are then named "obj" which is not useful when trying to find out which license is used in which project.
Thus in practice, the glob pattern is not useful with NuGet Source.

Suggestion: the source_path could be set to the project folder and NuGet Source internally checks for the "obj" subfolder (and if it does not exist, NuGet Source looks directly in the source_path for the project.assets.json). Then my glob would be `some/path/*` and the app name would be the project name.

Alternatively I could think of a new property just for the obj folder of an app.

Do you have any other suggestions?

jonabc:
@JanMattner interesting, when the NuGet source was first built I recall asking a question along similar lines as the above, and being told that the common pattern was to have a single `obj` folder outside the path of individual projects. The projects' obj outputs are subfolders of `obj` in that model - `some/path/obj/*` .

I'm not sure the best way to reconcile those two types of glob patterns, maybe a new `obj_path` configuration value or similar that the NuGet source can use on top of the given source paths?  so you could have something like the following to enable your use case?

```yaml
source_path: some/path/*
nuget:
  obj_path: obj
```

JanMattner:
That's interesting, I was not aware of that pattern and to be honest - I have never seen it in any project. At least the default behavior is one obj folder per project (csproj) and I think many Apps/Projects use it in such a way. So it would be great for GitHub Licensed to support this default pattern, so applying it is as easy as possible.

Great suggestion, with "obj_path" defaulting to an empty string, we could keep backwards compatibility to the current behavior.

JanMattner:
Any news on this?

jonabc:
@JanMattner sorry no, at least not from my perspective.  I've been busy with other tasks and don't have much bandwidth right now to look at this.  I could try to fit this into one of the next few weekends, but can't make any promises.  If you have a pressing need and are familiar with Ruby I'd happily review/advise on a PR though

JanMattner:
@jonabc Thanks for the PR! Unfortunately I have never done anything with Ruby :( 
But I can try to build and test it in a real world example.

PR #365, opened 2021-05-31
Add configurable property for nuget project obj path

jonabc:
closes https://github.com/github/licensed/issues/355

This adds an `obj_path` property for the nuget source that should make it easier/possible to specify nuget sources with a glob source path.  Specifically as mentioned in [this comment](https://github.com/github/licensed/issues/355#issuecomment-822647991), the nuget source can be configured with the following to enable glob patterns.

```
source_path: some/path/*
nuget:
  obj_path: obj
```

/cc @JanMattner FYI as issue author

/cc @zarenner 👋 as being involved with the original nuget source implementation.  It's been a _long_ time since I've worked with .Net and C#, but I do recall the default project structure for obj folders being `project/obj/...`.  I think it would make sense to make the above "obj" value the default for `obj_path` in the next breaking major version release, and allow this to be overridden for other project structures.  WDYT?

### DIFF
--- a/lib/licensed/sources/nuget.rb
+++ b/lib/licensed/sources/nuget.rb
@@ -164,12 +164,16 @@ module Licensed
       end
 
       def project_assets_file_path
-        File.join(config.pwd, "project.assets.json")
+        File.join(config.pwd, nuget_obj_path, "project.assets.json")
       end
 
       def project_assets_file
         return @project_assets_file if defined?(@project_assets_file)
         @project_assets_file = File.read(project_assets_file_path)
+      end
+
+      def nuget_obj_path
+        config.dig("nuget", "obj_path") || ""
       end
 
       def enabled?
--- a/test/sources/nuget_test.rb
+++ b/test/sources/nuget_test.rb
@@ -39,6 +39,18 @@ if Licensed::Shell.tool_available?("dotnet")
           assert_equal "https://www.newtonsoft.com/json", dep.record["homepage"]
         end
       end
+
+      it "finds dependendencies under a configured obj path" do
+        Dir.chdir File.join(fixtures, "..") do
+          config["nuget"] = { "obj_path" => "obj" }
+          dep = source.dependencies.detect { |d| d.name == "Newtonsoft.Json-12.0.3" }
+          assert dep
+          assert_equal "nuget", dep.record["type"]
+          assert_equal "Newtonsoft.Json", dep.record["name"]
+          assert_equal "12.0.3", dep.record["version"]
+          assert_equal "https://www.newtonsoft.com/json", dep.record["homepage"]
+        end
+      end
     end
 
     describe "license expressions" do


JanMattner:
Thanks for the PR!
I have tested it and it works quite well.
However, if I use a rather general glob pattern such as `mypath/**/*`, then it takes a while, because it scans all folders, even `bin` or `obj` or the `packages` folder - I can't see how I can exclude folders here (do I miss something?).
Furthermore, it lists all scanned / checked folders even those with 0 dependencies, so I get a lot of unnecessary output that makes it harder to actually see the relevant parts.

Is there any possibility to exclude certain folders or not to display the output for folders with 0 dependencies?

Example output for this folder structure:
image
with projects in `a1.b1`, `a1.b2` and `a2`:
```
C:\mytest>licensed cache
Caching dependency records for mytest-a1
Caching dependency records for mytest-a1.b1
  nuget
    Using Microsoft.Practices.Prism-4.1.0 (4.1.0)
    Using Microsoft.Practices.ServiceLocation-5.0.505 (5.0.505)
  * 2 nuget dependencies
Caching dependency records for mytest-a1.b2
  nuget
    Using Microsoft.Practices.Prism-4.1.0 (4.1.0)
    Using Microsoft.Practices.ServiceLocation-5.0.505 (5.0.505)
  * 2 nuget dependencies
Caching dependency records for mytest-a2
  nuget
    Using Microsoft.Practices.Prism-4.1.0 (4.1.0)
    Using Microsoft.Practices.ServiceLocation-5.0.505 (5.0.505)
  * 2 nuget dependencies
Caching dependency records for mytest-a3
Caching dependency records for mytest-bin
Caching dependency records for mytest-bin
Caching dependency records for mytest-bin
Caching dependency records for mytest-obj
Caching dependency records for mytest-obj
Caching dependency records for mytest-obj

C:\mytest>licensed status
Checking cached dependency records for mytest-a1

0 dependencies checked, 0 errors found.
Checking cached dependency records for mytest-a1.b1
FF
Errors:
* mytest-a1.b1.nuget.Microsoft.Practices.Prism-4.1.0
  filename: C:/mytest/.licenses/a1.b1/nuget/Microsoft.Practices.Prism-4.1.0.dep.yml
    - missing license text
    - license needs review: none

* mytest-a1.b1.nuget.Microsoft.Practices.ServiceLocation-5.0.505
  filename: C:/mytest/.licenses/a1.b1/nuget/Microsoft.Practices.ServiceLocation-5.0.505.dep.yml
    - missing license text
    - license needs review: none


2 dependencies checked, 4 errors found.
Checking cached dependency records for mytest-a1.b2
FF
Errors:
* mytest-a1.b2.nuget.Microsoft.Practices.Prism-4.1.0
  filename: C:/mytest/.licenses/a1.b2/nuget/Microsoft.Practices.Prism-4.1.0.dep.yml
    - missing license text
    - license needs review: none

* mytest-a1.b2.nuget.Microsoft.Practices.ServiceLocation-5.0.505
  filename: C:/mytest/.licenses/a1.b2/nuget/Microsoft.Practices.ServiceLocation-5.0.505.dep.yml
    - missing license text
    - license needs review: none


2 dependencies checked, 4 errors found.
Checking cached dependency records for mytest-a2
FF
Errors:
* mytest-a2.nuget.Microsoft.Practices.Prism-4.1.0
  filename: C:/mytest/.licenses/a2/nuget/Microsoft.Practices.Prism-4.1.0.dep.yml
    - missing license text
    - license needs review: none

* mytest-a2.nuget.Microsoft.Practices.ServiceLocation-5.0.505
  filename: C:/mytest/.licenses/a2/nuget/Microsoft.Practices.ServiceLocation-5.0.505.dep.yml
    - missing license text
    - license needs review: none


2 dependencies checked, 4 errors found.
Checking cached dependency records for mytest-a3

0 dependencies checked, 0 errors found.
Checking cached dependency records for mytest-bin

0 dependencies checked, 0 errors found.
Checking cached dependency records for mytest-bin

0 dependencies checked, 0 errors found.
Checking cached dependency records for mytest-bin

0 dependencies checked, 0 errors found.
Checking cached dependency records for mytest-obj

0 dependencies checked, 0 errors found.
Checking cached dependency records for mytest-obj

0 dependencies checked, 0 errors found.
Checking cached dependency records for mytest-obj

0 dependencies checked, 0 errors found.
```

jonabc:
@JanMattner would `mypath/*/*` be an ok glob, rather than `mypath/**/*`?  That would pick only the folders that are nested at the second level rather than finding all subfolders

JanMattner:
@jonabc For that specific project, it might work, becuase the csproj files are located at the same folder hierarchy level. But that is just by chance and that could change in the future.
In other projects the csproj files can be located in different folder depths.

Furthermore, a generic glob pattern has the advantage, that new added projects cannot be missed in the license check.

So the answer to your question: yes, that glob pattern could be used in this case, but in general a way to exclude folders would be great - at least "0 dependency" folders should be hidden in the output (maybe with an additional flag/parameter?)

JanMattner:
@jonabc I have another use case: I want to exclude test projects that e.g. end with ".Test"

With the normal glob, I cannot do that.
There are packages for "extended" globbing for the shell, e.g. https://unix.stackexchange.com/a/164026
I don't know, if something similar is available in Python.

jonabc:
> at least "0 dependency" folders should be hidden in the output (maybe with an additional flag/parameter?)
> I want to exclude test projects that e.g. end with ".Test"

Do test projects have a different designation someplace that is machine parseable?  Does the csproj or project.assets.json file denote a test project differently from a non-test project?

Can you list out the types of projects that are currently being evaluated, but you don't want them to be?  I'd much prefer to focus on project types specific to nuget rather than broader changes affecting licensed as a whole.

JanMattner:
@jonabc As far as I know, in the full / old .NET Framework (4.8 and older) there are project types defined by GUIDs, but in dotnet core I don't know anything similar.
Usually, each project sticks to some naming convention, that's why your new proposal in the PR #368 is I think the best solution - because each project may want to ignore various projects (e.g. internal, test, prototype, ...) that can always be identified by their name.

jonabc:
@JanMattner yeah I looked through all the documentation I could find along with the test project in the fixtures directory and couldn't find any fields that would make a distinction.

I generally try to avoid making changes to core logic for a single source enumerator if I can avoid it, but in this case it seems to be necessary :shrug:

jonabc:
merging this since the only feedback was resolved in https://github.com/github/licensed/pull/368

Issue #352, opened 2021-04-08
NuGet Source should search in fallbackFolders if package is not found in packagesPath


JanMattner:
Typically, the System/MS packages are not stored in the user's NuGet cache folder, but in the SDK's "Fallback Folder".

Part of my project.assets.json:

```
  "packageFolders": {
    "C:\\Users\\jan.mattner\\.nuget\\packages\\": {},
    "C:\\Program Files\\dotnet\\sdk\\NuGetFallbackFolder": {}
  },
  "project": {
    "version": "1.1.2289.7",
    "restore": {
      "projectUniqueName": "C:\\path\\to\\project\\myproject.csproj",
      "projectName": "MyProject",
      "projectPath": "C:\\path\\to\\project\\myproject.csproj",
      "packagesPath": "C:\\Users\\jan.mattner\\.nuget\\packages\\",
      "outputPath": "C:\\path\\to\\project\\obj\\",
      "projectStyle": "PackageReference",
      "crossTargeting": true,
      "fallbackFolders": [
        "C:\\Program Files\\dotnet\\sdk\\NuGetFallbackFolder"
      ],
```

Currently, the NuGet Source only considers:
https://github.com/github/licensed/blob/0d31c466dabaea259ccef45dd5ad823c56a29780/lib/licensed/sources/nuget.rb#L184

Then with the `licensed cache` command I get the following warning:

```
   * Warnings:
    * MyProject.nuget.Microsoft.CSharp-4.0.1
      - expected dependency path C:\Users\jan.mattner\.nuget\packages\microsoft.csharp/4.0.1 does not exist
```

But in fact, it exists at `C:\Program Files\dotnet\sdk\NuGetFallbackFolder\microsoft.csharp\4.0.1`

So I propose that if the file is not found in the "primary" packagesPath, then it should search all the other fallbackFolders.

jonabc:
:wave: Hi @JanMattner thanks for bringing this up!  There's a PR about this that's been open for quite awhile at https://github.com/github/licensed/pull/273, I'll take another look at that and see if I can clean that up to get it merge-able.

PR #366, opened 2021-05-31
Find packages from Nuget fallback folders if possible

jonabc:
closes https://github.com/github/licensed/issues/352

This gives an alternate solution to https://github.com/github/licensed/pull/273. I opted to start a new branch because that PR/branch had become stale and it was a bit quicker to build on the ideas rather than building directly on the code.

I've given attribution to @paveliak in the commit comment, and here as well I guess 😄 .  @paveliak if there's another way you'd like me to represent your input in these changes please let me know 🙏 !

Some other changes that came along as I better understood the json file format and the code here:
1. raising a `Licensed::Sources::Source::Error` error if the json file can't be parsed.  This is a special error class that signals a source enumerator-wide error.
2. setting a custom error message on the dependency if a path isn't found. a default error is set if a path is given but doesn't exist on disk, however in this case the path would be `nil`, which doesn't work with the default scenario.
3. I refactored the code around finding reference keys to (I hope) simplify the code readability just a little bit.  I'm using a `Set` to enforce uniqueness, though now the uniqueness is enforced for `reference_key` values and not computed `id` values.

### DIFF
--- a/lib/licensed/sources/nuget.rb
+++ b/lib/licensed/sources/nuget.rb
@@ -178,7 +178,7 @@ module Licensed
         message = "Licensed was unable to read the project.assets.json file. Error: #{e.message}"
         raise Licensed::Sources::Source::Error, message
       end
-      
+
       def nuget_obj_path
         config.dig("nuget", "obj_path") || ""
       end
--- a/lib/licensed/sources/nuget.rb
+++ b/lib/licensed/sources/nuget.rb
@@ -195,7 +195,7 @@ module Licensed
 
           path = full_dependency_path(reference_key)
           error = "Package #{id} not found at any project package folder" if path.nil?
-          
+
           NuGetDependency.new(
             name: id,
             version: version,
--- a/lib/licensed/sources/nuget.rb
+++ b/lib/licensed/sources/nuget.rb
@@ -198,7 +198,7 @@ module Licensed
           id = "#{name}-#{version}"
 
           path = full_dependency_path(reference_key)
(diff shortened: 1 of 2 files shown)


jonabc:
@zarenner :wave: can you review this PR?  I'm pretty confident it should be ok, but the sanity check on nuget usage and patterns as well as the extra set of eyes would be appreciated!

lib/licensed/sources/nuget.rb
```diff
@@ -184,32 +191,51 @@ def enabled?
...
-          end
-        end.values
+        reference_keys.map do |reference_key|
+          package_id_parts = reference_key.partition("/")
+          name = package_id_parts[0]
+          version = package_id_parts[-1]
+          id = "#{name}-#{version}"
+
+          path = full_dependency_path(reference_key)
+          error = "Package #{id} not found at any project package folder" if path.nil?
```
zarenner, line 201: Won't `path` also be `nil` if for some reason it's missing under libraries -> `reference_key` -> path? Not a big deal but the error will be slightly incorrect.

zarenner (approved):

zarenner:
lgtm!

lib/licensed/sources/nuget.rb
zarenner: (feel free to ignore, I don't think this is important enough to differentiate on, but perhaps you want to clarify the message)

jonabc: 👍  good catch

PR #367, opened 2021-05-31
Add available metadata to swift dependencies

jonabc:
:wave: @mattt it's been awhile since a new dependency source has been added to licensed and I missed that there is some metadata content that is nice to include if/when it's available.

"type" is another indicator of the source of the dependency.  This can also be found from the path to the cached metadata, but 🤷 including it here will put that value into the cached file for a little added helpfulness with minimal overhead.

"homepage" is useful for someone to track the dependency back to its source, and in this case should _usually_ be available from the git url.  It's not a 100% guarantee that taking ".git" off of the "repositoryURL" field will result in a valid URL, however there's a pretty good chance it'll work and it's a low cost addition.

The last piece of common metadata that is included if it's available is a summary or description of the dependency package.  It doesn't look like that's available from the information here, at least from the test project.  Is a description included in any distributed content in swift's package management?

Can you take a look and let me know if you see anything off or wrong about these changes?

### DIFF
--- a/lib/licensed/sources/swift.rb
+++ b/lib/licensed/sources/swift.rb
@@ -15,7 +15,8 @@ module Licensed
         pins.map { |pin|
           name = pin["package"]
           version = pin.dig("state", "version")
-          url = pin["repositoryURL"].sub(/\.git$/, "")
+          url = pin["repositoryURL"]
+          homepage = url.sub(/\.git$/, "")
           path = nil
           errors = []
 
@@ -32,7 +33,7 @@ module Licensed
             errors: errors,
             metadata: {
               "type"      => Swift.type,
-              "homepage"  => url
+              "homepage"  => homepage
             }
           )
         }
--- a/lib/licensed/sources/swift.rb
+++ b/lib/licensed/sources/swift.rb
@@ -54,7 +54,7 @@ module Licensed
       end
 
       def dependency_path_for_url(url)
-        last_path_component = URI(url).path.split("/").last
+        last_path_component = URI(url).path.split("/").last.sub(/\.git$/, "")
         File.join(config.pwd, ".build", "checkouts", last_path_component)
       end
 
--- a/lib/licensed/sources/swift.rb
+++ b/lib/licensed/sources/swift.rb
@@ -15,11 +15,12 @@ module Licensed
         pins.map { |pin|
           name = pin["package"]
           version = pin.dig("state", "version")
+          url = pin["repositoryURL"].sub(/\.git$/, "")
           path = nil
           errors = []
 
           begin
-            path = dependency_path_for_url(pin["repositoryURL"])
+            path = dependency_path_for_url(url)
           rescue => e
             errors << e
           end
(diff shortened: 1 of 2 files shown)


lib/licensed/sources/swift.rb
```diff
@@ -48,7 +53,7 @@ def pins
       end
 
       def dependency_path_for_url(url)
-        last_path_component = URI(url).path.split("/").last.sub(/\.git$/, "")
+        last_path_component = URI(url).path.split("/").last
```
mattt, line 56: When Swift Package Manager clones Git repositories, it follows the same default behavior as the `git` client, where the resulting directory is named with the basename of the last path component. Wouldn't this change produce the wrong path for the cloned dependency?

mattt, line 56: When Swift Package Manager clones Git repositories, it follows the same default behavior as the `git` client, where the resulting directory is named with the basename of the last path component. That's the behavior I intended to capture in this method.

Any objections to keeping this as-is, and making an explicit call to `.sub(/\.git$/, "")` in `#enumerate_dependencies`?

mattt (approved):
Thanks for circling back on this, @jonabc. This looks like a good change to me. Just a small comment on my end. Feel free to apply the batch of suggestions or ignore.

lib/licensed/sources/swift.rb
```diff
@@ -15,11 +15,12 @@ def enumerate_dependencies
         pins.map { |pin|
           name = pin["package"]
           version = pin.dig("state", "version")
+          url = pin["repositoryURL"].sub(/\.git$/, "")
```
mattt, line 18:
```suggestion
          url = pin["repositoryURL"]
          homepage = url.sub(/\.git$/, "")
```

Although not as common, Git URLs can be specified with non-HTTP schemes. A more robust solution would be to implement a `homepage_for_url` helper method that returns `nil` in those cases.

lib/licensed/sources/swift.rb
```diff
@@ -28,7 +29,11 @@ def enumerate_dependencies
             name: name,
             path: path,
             version: version,
-            errors: errors
+            errors: errors,
+            metadata: {
+              "type"      => Swift.type,
+              "homepage"  => url
```
mattt, line 35:
```suggestion
              "homepage"  => homepage
```

lib/licensed/sources/swift.rb
jonabc: Ah yeah I see what you mean, I had moved it around because it seemed unnecessary to do that operation twice, and it was still matching the intended behavior.  I don't have a strong feeling here and happy to change it back.

jonabc:
> A more robust solution would be to implement a homepage_for_url helper method that returns nil in those cases.

@mattt yep, that's what I called out in the PR writeup 😄 
> "homepage" is useful for someone to track the dependency back to its source, and in this case should usually be available from the git url. It's not a 100% guarantee that taking ".git" off of the "repositoryURL" field will result in a valid URL, however there's a pretty good chance it'll work and it's a low cost addition.

That said, if you're looking for a better solution I updated the code to check for an http or https scheme as suggested 👍

mattt:
@jonabc Apologies for not responding sooner. I just saw that you merged this in. Thanks again for all of your help!

PR #368, opened 2021-06-06
Allow arrays of source paths

jonabc:
This PR addresses some comments that came up in https://github.com/github/licensed/pull/365 by adding gitignore-like exclusion glob patterns that can be used to better filter what paths licensed will evaluate.

To do this I applied the inclusion/exclusion pattern functionality that exists when specifying a manifest source configuration through a series of globs in the licensed configuration file.  This was used during configuration source path expansion, and surrounding code was adapted to keep everything working and to handle edge/error cases.

  One side effect of this is that it is now possible to specify an array with multiple non-glob paths as well.  This is already accomplishable by creating multiple `apps` in the configuration file.  There are potentially different use cases for specifying multiple source paths vs multiple apps, but it adds some confusion to the developer experience when determining how to configure licensed for multiple sources.  In order to minimize any confusion, I've only documented the use of arrays when using glob patterns.  This can of course be revisited in the future.

/cc @JanMattner does adding negative glob patterns to filter the set of app sources solve your use case?

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -19,9 +19,13 @@ If a root path is not specified, it will default to using the following, in orde
 1. the root of the local git repository, if run inside a git repository
 2. the current directory
 
-### Source path glob patterns
+### Source paths
 
-The `source_path` property can use a glob path to share configuration properties across multiple application entrypoints.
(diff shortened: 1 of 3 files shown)


JanMattner:
@jonabc Yes, thanks! I think that should solve all use cases I had in mind.

PR #373, opened 2021-06-16
Release 3.1.0

jonabc:
## 3.1.0

2021-06-16

### Added

- Licensed supports Swift/Swift package manager as a dependency source (:tada: @mattt https://github.com/github/licensed/pull/363)'

### Changed

- The `source_path` configuration property accepts arrays of inclusion and exclusion glob patterns (https://github.com/github/licensed/pull/368)
- The Nuget source now uses configured fallback folders to find dependencies that are not in found in the project folder (https://github.com/github/licensed/pull/366)
- The Nuget source supports a configurable property for the path from the project source path to the project's `obj` folder (https://github.com/github/licensed/pull/365)

### Fixed
- The Go source's checks for local packages will correctly find paths in case-insensitive file systems (https://github.com/github/licensed/pull/370)
- The Bundler source will no longer unnecessarily reset the local Bundler environment configuration (https://github.com/github/licensed/pull/372)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,24 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## 3.1.0
+
+2021-06-16
+
+### Added
+
+- Licensed supports Swift/Swift package manager as a dependency source (:tada: @mattt https://github.com/github/licensed/pull/363)'
+
+### Changed
+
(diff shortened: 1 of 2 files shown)


jonabc:
I'm not sure why bundler 1.15.0 CI is failing here but wasn't failing previously 🤷 .  I narrowed down the cause to a change in the logic on `::Bundler.reset!` in 1.16.0.  In order to have consistent behavior across versions I'd need to start mucking with instance variables on the Bundler module, so instead I think it might be time to update the supported Bundler minor versions (based on whats run in CI) to drop 1.15 and add 2.1, 2.2

PR #385, opened 2021-08-20
chore(ci) : update dependencies and define caches

mmorel-35:
- setup dependabot
- update current dependencies
- define caches for cabal, go , gradle, pip, spm


jonabc:
:wave: @mmorel-35 thanks for opening this.  Adding dependabot is 👍  in general, though I'm curious if there is any context around this update since the PR body is pretty sparse.

Were you finding any issues with the tool as-built?  Or is this preventative for future updates?

mmorel-35:
Hi ! There was no issue. I like the tool. I just want to help the project to spare time in the developement. With the bot you just validate he PR and with the cache you have you can do more builds in the same time .

.github/workflows/test.yml
```diff
@@ -339,10 +402,20 @@ jobs:
       with:
         ruby-version: 2.6
     - run: bundle lock
-    - uses: actions/cache@v1
+    - uses: actions/cache@v2
+      name: cache pip dependencies
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
```
jonabc, line 409: same here - this might need to include `${{ matrix.python }}` as part of the `key` and restore key

.github/workflows/test.yml
```diff
@@ -395,10 +471,20 @@ jobs:
       with:
         ruby-version: 2.6
     - run: bundle lock
-    - uses: actions/cache@v1
+    - uses: actions/cache@v2
+      name: cache spm dependencies
+      with:
+        path: .build
+        key: ${{ runner.os }}-spm-${{ hashFiles('**/Package.resolved') }}
```
jonabc, line 478: and same here with `${{ matrix.swift }}`

jonabc (commented):
thanks for making this change - I left a few requests to have greater specificity in the cache keys.  This tool's entire purpose is to work across a range of versions on a range of package managers and languages, and I've been surprised in the past at things going wrong between different versions of the language or package manager being used.

I'm similarly wary of the use of the restore keys that can potentially cause dangerous cache sharing.  I didn't add them in the past based on the same ☝️ concerns.  It's probably fine to try out and if re-evaluate if anything comes up though 🤷

.github/workflows/test.yml
```diff
@@ -211,10 +250,22 @@ jobs:
...
         java-version: ${{ matrix.java }}
         distribution: adopt
-    - uses: actions/cache@v1
+    - uses: actions/cache@v2
+      name: cache gradle dependencies
+      with:
+        path: |
+          ~/.gradle/caches
+          ~/.gradle/wrapper
+        key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
```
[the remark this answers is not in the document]
mmorel-35: If there is a different version of a dependency for a different version of java, then the files shall also be differents and so will the hash.

mmorel-35:
Hi! I will follow your préconisations.
By the way, I took examples from actions/cache. https://github.com/actions/cache/blob/main/examples.md 
If they missed something then it might worth helping them complete it in the process.

jonabc:
> If they missed something then it might worth helping them complete it in the process.

`actions/cache` and actions workflows in general provide the configurability that licensed needs.  TBH I don't have any specific requests from the `actions/cache` owners to make the usage better here. I'm guessing this tool falls into a smaller class of tools and applications that needs to test across a number of different versions of dependent tools and frameworks beyond what the tool itself is built on.
 
 Is there anything specific you had in mind that was missing?

mmorel-35:
I was thinking of documentation completion. For example with the cabal cache, there is no hashed file in the documentation as it can be seen for many other languages and tools

jonabc:
Ah got it.  Maybe that's something you'd like to open a PR on the actions/cache repo to update?  In any event, this change is looking good, thanks for opening this up and keeping things running smoothly 🙇

jonabc (approved):
:shipit: thanks!

PR #389, opened 2021-08-23
Update rubocop requirement from ~> 0.49, < 0.67 to ~> 0.49, < 1.20

mmorel-35:
Hi @jonabc , I have fixed the issue raised by dependabot's PR.
I have replaced XXHash64 with SHA256
Replaces #388


lib/licensed/sources/helpers/content_versioning.rb
```diff
@@ -63,9 +63,8 @@ def contents_hash(paths)
         return if paths.empty?
 
         paths.sort
-             .reduce(Digest::XXHash64.new, :file)
-             .digest
-             .to_s(16) # convert to hex
```
jonabc, line 68: this hash isn't used for any sort of security purposes, it's a fingerprint mark used to version a dependency based on the dependency's file contents.  Performance is more important than security here and XXHash64 is a faster algorithm in general than SHA256

Given it's usage, I'm not concerned about fixing the rubocop warning but I am concerned about the backwards compatibility concerns and churn around changing the algorithm used here.  I'd prefer if the cop raising the violation was [disabled around these lines](https://docs.rubocop.org/rubocop/1.19/configuration.html#disabling-cops-within-source-code) rather than changing the algorithm

jonabc (approved):
✨ thanks for fixing the failures and cleaning things up, much appreciated!

Issue #362, opened 2021-05-18
Could not find nokogiri (1.11.4) in any sources


tommeier:
```
  * Errors:
    * native.bundler.nokogiri
      - could not find nokogiri (1.11.4) in any sources
```

on: `v3.0.1`

Is working with same Gemfile on `2.14.4`

jonabc:
@tommeier :wave: can you provide a repro project or gemfile so I can take a better look at whats going on? 🙇   The bundler source went through some significant changes 2.x.x -> 3.x.x so I'm not terribly surprised to see some things like this pop up

tommeier:
hi @jonabc ,

Well - I'd love too... 😆  I'm just for some reason unable too. Completely baffled.

In our (albeit gigantic monolith) project, I get this every time with the root `Gemfile` simplified down to this:

```
source "https://rubygems.org"

group :development do
  gem "licensed", group: :development
end
gem 'appium_lib'
```

WIth the most simple `.licensed.yml` of:

```
cache_path: license_cache

sources:
  bundler: true
```

However, on a clean seperate project (same bundler `2.2.11` and same ruby `2.7.2`) this issue doesn't exist.

On our project root:

```
Caching dependency records for up
  bundler
    Using appium_lib (11.2.0)
    Using appium_lib_core (4.5.0)
    Using childprocess (3.0.0)
    Using eventmachine (1.2.7)
    Using faye-websocket (0.11.0)
    Using rubyzip (2.3.0)
    Using selenium-webdriver (3.142.7)
    Using tomlrb (1.3.0)
    Using websocket-driver (0.7.3)
    Using websocket-extensions (0.1.5)

  * Errors:
    * up.bundler.nokogiri
      - could not find nokogiri (1.11.4) in any sources
```

On a clean repo with the same gemfile + licensed.yml:

```
Caching dependency records for test-bundle
  bundler
    Using appium_lib (11.2.0)
    Using appium_lib_core (4.5.0)
    Using childprocess (3.0.0)
    Using eventmachine (1.2.7)
    Using faye-websocket (0.11.0)
    Using nokogiri (1.11.4)
    Using racc (1.5.2)
    Using rubyzip (2.3.0)
    Using selenium-webdriver (3.142.7)
    Using tomlrb (1.3.0)
    Using websocket-driver (0.7.3)
    Using websocket-extensions (0.1.5)
  * 12 bundler dependencies
```


[...]

PR #392, opened 2021-09-04
Remove more custom bundler logic and reload the bundler runtime for each app

jonabc:
closes https://github.com/github/licensed/issues/362

This change primarily looks to fix the linked issue, which is done by reloading the bundler runtime with `::Bundler.load` after resetting bundler's configuration with `::Bundler.reset!`.  This should force bundler to look at the current directory's gem and lock files in determining things like available platforms.

I found that with this change, Bundler acted like `licensed` was being run from the context of the project being evaluated.  This allowed me to further pare down much of the custom logic that had been built up on the bundler source enumerator over time.  The bundler source has been a love/hate relationship but I'm finding now that the best option usually involves not trying to override Bundler's behavior, or when necessary making very targeted functionality changes.

This PR does include such changes, in that it monkey patches `::Bundler::Definition#specs` and `::Bundler::Definition.requested_groups`. This simplifies the logic that had been built around managing which specs were or were not evaluated by licensed.  With the monkey patch it's become easier to make targeted changes to behavior due to operating in the context of the definition object and having direct access to private methods


tommeier:
Fixes it perfectly!! Thank you!! 😍 

👍

PR #400, opened 2021-09-12
ci(workflow): run core first

mmorel-35:
Run core first to benefits from it's cache in case of success or fail fast in case of failure on the core


jonabc (approved):
sure, lets give it a shot 👍

PR #411, opened 2021-09-30
Don't error when an indirect peerDependency is missing in npm 7+

manuelpuyol:
### Problem

The information provided by npm changed in v7+. E.g:

Before:

```json
"@react-native-community/async-storage": {
  "_id": "@react-native-community/async-storage@^1.2.0",
  "name": "@react-native-community/async-storage",
  "version": "^1.2.0",
  "peerMissing": [
    {
      "requiredBy": "@optimizely/js-sdk-datafile-manager@0.5.0",
      "requires": "@react-native-community/async-storage@^1.2.0"
    }
  ],
  "dependencies": {}
}
```

After:

```json
"@react-native-community/async-storage": {
  "missing": true,
  "problems": [
    "missing: @react-native-community/async-storage@^1.2.0, required by @optimizely/js-sdk-datafile-manager@0.5.0"
  ]
}
```

The main issue is the removal of the `peerMissing` key, which indicated that the dependency could be ignored.

**Edit:**

Another issue is that when a dependency is missing, npm7 does not add the `version` key, which was expected in npm < 7.
To work around this issue I made a similar fix, where I look for the parent `_dependencies` and `peerDependencies` to figure out the expected version.

### Proposed solution

To reproduce the `peerMissing` behavior, I'm updating `recursive_dependencies` to receive the parent dependency. When a dependency is `missing`, we can check if it was a `peerDependency` on the parent, which _should_ be the same as `peerMissing`.


lib/licensed/sources/npm.rb
```diff
@@ -135,6 +135,14 @@ def yarn_lock_present
       def include_non_production?
         config.dig("npm", "production_only") == false
       end
+
+      def missing_peer?(parent, dependency, name)
+        dependency["peerMissing"] || dependency["missing"] && peer_dependency?(parent, name)
+      end
+
+      def peer_dependency?(parent, name)
+        parent["peerDependencies"] && parent["peerDependencies"][name]
```
jonabc, line 144: nit: a bit shorter syntax here with [dig](https://ruby-doc.org/core-3.0.2/Hash.html#method-i-dig)
```suggestion
        parent.dig("peerDependencies", name)
```

lib/licensed/sources/npm.rb
```diff
@@ -135,6 +135,14 @@ def yarn_lock_present
       def include_non_production?
         config.dig("npm", "production_only") == false
       end
+
+      def missing_peer?(parent, dependency, name)
+        dependency["peerMissing"] || dependency["missing"] && peer_dependency?(parent, name)
```
jonabc, line 140: I _think_ you'd want parentheses around the last two checks right?  `peer_dependency?` would only be called as additional validation on `dependency["missing"]`?
```suggestion
        dependency["peerMissing"] || (dependency["missing"] && peer_dependency?(parent, name))
```

jonabc (commented):
I'm not sure what's going on with CI, it's possible though that the latest node 16 / npm 7 versions have fixed or otherwise changed logic that I had been working around with that test.

manuelpuyol:
> I'm not sure what's going on with CI, it's possible though that the latest node 16 / npm 7 versions have fixed or otherwise changed logic that I had been working around with that test.

I'm pretty confused with what's happening in CI 😅 

I added `"@optimizely/optimizely-sdk": "4.0.0",` to the npm7 `package.json` to really test the `peer` behavior, but for some reason `@react-native-community/async-storage` is also being installed and something is happening with `glob` and `graceful-fs`.

manuelpuyol:
Looks like [npm7 automatically installs peer dependencies by default](https://github.blog/2021-02-02-npm-7-is-now-generally-available/#peer-dependencies). So I added the `--legacy-peer-deps` flags when installing using npm7 (let me know if this makes sense or if we should just skip that test).

Since the package.json files look the same now I decided to unify them and simplify `script/source-setup/npm`.

There was a version error in a test which I addressed in the `Edit` provided in the PR description.

script/source-setup/npm
```diff
@@ -15,20 +15,14 @@ if [ "$1" == "-f" ]; then
...
 fi
 
-npm install
+NPM_MAJOR_VERSION="$(npm -v | cut -d'.' -f1)"
+if [ "$NPM_MAJOR_VERSION" -ge "7" ]; then
+  # do no install peerDependencies in npm 7
+  npm install --legacy-peer-deps
+else
+  npm install
+fi
```
jonabc, line 28: I was curious about what worked or didn't work when installing with the flag or not and checked out the branch locally to play around with things.  It looks like all the tests pass as expected whether the flag is set or not 🤷 .  Do you know of any reason why installing with the flag is needed or can it be reverted?

lib/licensed/sources/npm.rb
```diff
@@ -135,6 +137,18 @@ def yarn_lock_present
...
+        dependency["peerMissing"] || (dependency["missing"] && peer_dependency(parent, name))
+      end
+
+      def peer_dependency(parent, name)
+        parent.dig("peerDependencies", name)
+      end
+
+      def extract_version(parent, name)
+        parent.dig("_dependencies", name) || peer_dependency(parent, name)
+      end
```
jonabc, line 151: can you add comments about the expected return values from each of these functions?  It was a little confusing that `peer_dependency` returned a usable version value, maybe that function could also be renamed?

lib/licensed/sources/npm.rb
```diff
@@ -135,6 +137,18 @@ def yarn_lock_present
       def include_non_production?
         config.dig("npm", "production_only") == false
       end
+
+      def missing_peer?(parent, dependency, name)
+        dependency["peerMissing"] || (dependency["missing"] && peer_dependency(parent, name))
+      end
+
+      def peer_dependency(parent, name)
+        parent.dig("peerDependencies", name)
```
jonabc, line 146: it's possible for parent to be nil if this gets called from the `recursive_dependencies` call on the root dependencies at line 54. It's probably a good idea to update that call to include `package_metadata` as the parent, but for extra safety this should use the nil safe operator

```suggestion
        parent&.dig("peerDependencies", name)
```

lib/licensed/sources/npm.rb
```diff
@@ -135,6 +137,18 @@ def yarn_lock_present
...
+      end
+
+      def extract_version(parent, name)
+        parent.dig("_dependencies", name) || peer_dependency(parent, name)
```
jonabc, line 150: it's possible for parent to be nil if this gets called from the recursive_dependencies call on the root dependencies at line 54. It's probably a good idea to update that call to include package_metadata as the parent, but for extra safety this should use the nil safe operator
```suggestion
        parent&.dig("_dependencies", name) || peer_dependency(parent, name)
```

jonabc (commented):
This looks great, thanks for checking on peer dependency installation in npm 7 🙇 !  A few more small change requests, nothing big. Thanks again for putting this together 🎉

script/source-setup/npm
manuelpuyol: huh, when I installed locally with npm7 (fresh setup) without the flag it also installed the `@react-native-community/async-storage` dependency, which failed some tests
I'm going to try again and see what happens

jonabc: huh I must have put things into a weird state locally 🤷 .  shouldn't block this PR, happy to leave this as-is and I can figure it out later

lib/licensed/sources/npm.rb
jonabc: @manuelpuyol you resolved this comment but the possibility of a nil reference still exists if parent is nil.  Note this was a second comment of this type, separate from the change you already committed

manuelpuyol: must have been a missclick, sorry 😅

jonabc (approved):
❤️  thanks for this change and working through the feedback

lib/licensed/sources/npm.rb
villelahdenvuo: @jonabc Looks like parent can be a String sometimes... Weird.

image

jonabc:
@villelahdenvuo :wave: I saw you commented about a bug but I'm not seeing it here and it looks like you might have deleted the comment.  Were you able to get the issue resolved?  If so can you share what happened?  Is there additional error handling that could be added when evaluating npm sources?

villelahdenvuo:
Hey @jonabc It's in the resolved comments, here it is:

image

I got it working with 3.3.0, I don't really know what the root cause is. I don't have a repro because it's in a private repo, but I guess it has something to do with peerDependencies, but our project doesn't have any.

Issue #418, opened 2021-11-17
Should the license cache be committed to version control?


andreas-hakansson-ingka:
Is the intended workflow that the license cache should be committed to version control, with the rest of the project? I'm guessing yes? But I can't see the folder in this repository.

andreas-hakansson-ingka:
@jonabc thanks for the update. We're running into a bit of a problem, in some of our repositorsies though. We're using Licensed to scan go dependencies. Locally we use macOS for development, but we also run Licensed as part of our GitHub actions.

It seems that there are some slight differences in the dependencies that could be generated for Go depending on the host operating system. For example, if we run `licensed cache` locally then part of the cached dependencies are

```sh
    Caching golang.org/x/net/context/ctxhttp (v0.0.0-20210510120150-4163338589ed)
    Caching golang.org/x/net/http2 (v0.0.0-20210510120150-4163338589ed)
    Caching golang.org/x/net/internal/timeseries (v0.0.0-20210510120150-4163338589ed)
    Caching golang.org/x/net/trace (v0.0.0-20210510120150-4163338589ed)
    Caching golang.org/x/oauth2 (v0.0.0-20210514164344-f6687ab2804c)
    Caching golang.org/x/oauth2/authhandler (v0.0.0-20210514164344-f6687ab2804c)
    Caching golang.org/x/oauth2/google (v0.0.0-20210514164344-f6687ab2804c)
    Caching golang.org/x/oauth2/google/internal/externalaccount (v0.0.0-20210514164344-f6687ab2804c)
    Caching golang.org/x/oauth2/internal (v0.0.0-20210514164344-f6687ab2804c)
    Caching golang.org/x/oauth2/jws (v0.0.0-20210514164344-f6687ab2804c)
    Caching golang.org/x/oauth2/jwt (v0.0.0-20210514164344-f6687ab2804c)
    Caching golang.org/x/sync/semaphore (v0.0.0-20210220032951-036812b2e83c)
```

However, if we were to use this cache in GitHub Actions (running on an Ubuntu operating system) then out `licensed status` would fail with

```sh
* REDACTED.go.golang.org/x/sys/internal/unsafeheader
  filename: /home/runner/work/REDACTED/.licenses/go/golang.org/x/sys/internal/unsafeheader.dep.yml, version: v0.0.0-20210514084401-e8d321eab015, license: , allowed: false
    - cached dependency record not found

make: *** [Makefile:85: licensed-status] Error 1
[...]

PR #420, opened 2021-11-24
Add license caching action workflow

jonabc:
ref https://github.com/github/licensed/issues/418

This adds a workflow to the repo that uses https://github.com/jonabc/licensed-ci to keep cached license files up to date.  I've added comments around most configuration of the workflow because this might be the first place that someone sees the action and it's helpful to understand what it's doing 😄


jonabc:
🤦 closing this and opening a new PR on a different branch.  the branch name matches the pattern expected for a license update branch using the `branch` licensed-ci workflow, so it's not attempting to update cached files

PR #431, opened 2022-01-07
Don't crash yarn source on malformed json

jonabc:
closes https://github.com/github/licensed/issues/430

This is a minimal fix to resolve the crash.  In the linked issue the problem came from a dependency shipping it's intentionally malformed test fixture, which is not required and fine to ignore.  If there is an error when parsing a required package.json file with this change, a dependency will show up as not being found rather than the true error that package.json file couldn't be parsed.  That can be handled at a later time though

cc @pfeffer


jonabc:
The `licensed-ci` GitHub Action has created a [pull request](https://github.com/github/licensed/pull/432) containing license metadata updates based on the changes in this branch.

PR #443, opened 2022-01-31
Handle additional case of missing optional npm peer dependencies

jonabc:
Found this in a [dependabot update to licensed-ci](https://github.com/jonabc/licensed-ci/pull/115), it looks like recent versions of npm use a new method to signal for optional peer dependencies.  The `npm list` output I'm seeing here looks like the following.

```json
"node-fetch": {
      ...
      "peerDependencies": {
        "encoding": "^0.1.0"
      },
      "peerDependenciesMeta": {
        "encoding": {
          "optional": true
        }
      },
      ...
      "dependencies": {
        "encoding": {},
        ...
```

This PR adds a check for the combination of an empty dependency hash, being set in `peerDependencies`, and having optional set in `peerDependenciesMeta`.

### DIFF
--- a/.licenses/bundler/bundler.dep.yml
+++ b/.licenses/bundler/bundler.dep.yml
@@ -1,6 +1,6 @@
 ---
 name: bundler
-version: 2.3.5
+version: 2.3.6
 type: bundler
 summary: The best way to manage your application's dependencies
 homepage: https://bundler.io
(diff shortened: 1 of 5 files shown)


jonabc:
The `licensed-ci` GitHub Action has created a [pull request](https://github.com/github/licensed/pull/444) containing license metadata updates based on the changes in this branch.

Issue #448, opened 2022-02-03
Licensed crashing evaluating NPM sources


jonabc:
From https://github.com/github/licensed/pull/411#discussion_r798397392

![img](https://user-images.githubusercontent.com/633524/152320824-4ffe7f80-8b51-41ae-85ff-48b18ce905aa.png)
>  Looks like parent can be a String sometimes... Weird.

@villelahdenvuo FYI I'm opening a new issue on this so I can track it towards a fix.

You mentioned you can't provide a repro since it's coming from a private repo.  I don't need the full repo to reproduce the issue, only the package.json/package-lock.json files.  Would it be possible to share those if they don't contain sensitive/private/proprietary information?  Or is there a minimal set of npm packages you're seeing the issue with that doesn't include anything private?

I can "fix" the crash, but without knowing how or why the value is a string I'm concerned about any fix introducing a new bug.

jonabc:
I'm fixing this in https://github.com/github/licensed/pull/450.  A `peerDependencies` property unexpectedly had a string value, where licensed was expecting a hash.

jonabc:
the fix for this is included in [release 3.4.4](https://github.com/github/licensed/releases/tag/3.4.4).  Thanks again for the report and helping track down the cause

PR #450, opened 2022-02-07
Protect from crashes when using Hash#dig

jonabc:
resolves https://github.com/github/licensed/issues/448 https://github.com/github/licensed/issues/449

This PR makes two changes to protect against failures reported `String does not have #dig method (TypeError)`

1. In the npm source, it's possible for a `peerDependencies` value to be set as `"[Circular]"`.  I've only been able to reproduce this on npm 6.x.  This was tricky to track down as the error is being thrown internally from Hash#dig.  From the line of code in licensed `parent&.dig("peerDependencies", name)`, parent here is a hash but `parent["peerDependencies"]` is a string.  If I had to guess, I'd bet that the dig method is implemented using recursion like:
```ruby
def dig(*parts)
  key = parts[0]
  value = self[key]
  # no more key path parts to evaluate or value is nil, complete recursion
  return value if parts.length == 1 || value == nil

  # call dig on remaining key path parts
  # this is not resilient to the value not being a hash
  # and most likely value is not previously tested with respond_to?(:dig)
  value.dig(parts[1..])
end
```

2. A safety check in the pip source when using dig to evaluate the python config.  There is an expectation that a user has set the `python` key in the configuration file to a YAML hash, but in this case it looks like it was set to a string

### DIFF
--- a/docs/sources/pip.md
+++ b/docs/sources/pip.md
@@ -20,5 +20,5 @@ You have to add this setting to your licensed configuration file.
 An example usage of this might look like:
 ```yaml
 python:
-    virtual_env_dir:"/path/to/your/venv_dir"
+    virtual_env_dir: "/path/to/your/venv_dir"
 ```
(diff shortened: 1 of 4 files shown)


jonabc:
I put up a change to the npm test fixture on it's own that I used to reproduce the failure case for the npm source.  It should show the failure, where a successful CI run from a commit including the fix will validate the fix.

rwwagner90:
I'm still getting this error with 4.0.4.

PR #455, opened 2022-02-13
Allow custom licensee behavior overrides

jonabc:
closes https://github.com/github/licensed/issues/453

cc @villelahdenvuo :wave: this turned out to be really straightforward to implement so I went ahead and made the change rather than putting details in https://github.com/github/licensed/issues/453

This change allows users to [customize licensee's behavior](https://github.com/licensee/licensee/blob/jonabc-patch-1/docs/customizing.md) based on configuration set in the licensed configuration file.  Currently the only allowed customization is on the confidence threshold, however it could make sense to allow users to disable matching licenses from readme and package manager files if they choose to do so 🤷 

Licensee supports changing the confidence threshold via a global setting `Licensee.confidence_threshold`.  I'm setting (and restoring) the global value when evaluating each app config for the `licensed cache` command.  I don't believe any other licensed commands are affected by customizing licensee.  Setting the threshold per app config allows users to make use of licensed's inherited configuration and overrides to change the value for all, some, or only one configured app.

I've added documentation for customizing licensee behavior, however I'm intentionally not adding an example to the [common configuration values example](https://github.com/github/licensed/tree/master/docs/configuration.md).  Changing this value should be done intentionally, and I don't want anyone setting up a new project that might copy/paste the example configuration shown on that page to inadvertently customize licensee's behavior.

cc @mlinksva @benbalter FYI on this change to licensed<>licensee interactions

### DIFF
--- a/docs/configuration/customizing_licensee.md
+++ b/docs/configuration/customizing_licensee.md
@@ -1,0 +1,13 @@
+# Customize Licensee's behavior
+
+Licensed uses [Licensee](https://github.com/licensee/licensee) to detect and evaluate OSS licenses for project dependencies found during source enumeration.  Licensed can optionally [customize Licensee's behavior](https://github.com/licensee/licensee/blob/jonabc-patch-1/docs/customizing.md#customizing-licensees-behavior) based on options set in the configuration file.
+
+**NOTE** Matching licenses based on package manager metadata and README references is always enabled and cannot currently be configured.
+
+```yml
+licensee:
+  # the confidence threshold is an integer between 1 and 100. the value represents
+  # the minimum percentage confidence that Licensee must have to report a matched license
+  # https://github.com/licensee/licensee/blob/jonabc-patch-1/docs/customizing.md#adjusting-the-confidence-threshold
+  confidence_threshold: 90 # default value: 98
+```
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,7 +23,7 @@ jobs:
     steps:
       - id: get_version
         name: Get package version
-        uses: actions/github-script@v5
+        uses: actions/github-script@v6
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           result-encoding: string
@@ -41,7 +41,7 @@ jobs:
 
       - id: get_url
         name: Get release upload url
-        uses: actions/github-script@v5
+        uses: actions/github-script@v6
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           result-encoding: string
@@ -65,7 +65,7 @@ jobs:
 
       - id: get_ref
         name: Get checkout ref for custom build scripts
-        uses: actions/github-script@v5
+        uses: actions/github-script@v6
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           result-encoding: string
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -347,7 +347,7 @@ jobs:
         elixir: [ 1.11.x, 1.12.x ]
     steps:
     - uses: actions/checkout@v2
-    - uses: erlef/setup-elixir@v1.10
+    - uses: erlef/setup-elixir@v1.11
       with:
         otp-version: ${{matrix.otp}}
         elixir-version: ${{matrix.elixir}}
--- a/.licenses/bundler/bundler.dep.yml
+++ b/.licenses/bundler/bundler.dep.yml
@@ -1,6 +1,6 @@
 ---
 name: bundler
-version: 2.3.7
+version: 2.3.8
 type: bundler
 summary: The best way to manage your application's dependencies
 homepage: https://bundler.io
--- a/.licenses/bundler/bundler.dep.yml
+++ b/.licenses/bundler/bundler.dep.yml
@@ -1,6 +1,6 @@
 ---
 name: bundler
-version: 2.3.6
+version: 2.3.7
 type: bundler
 summary: The best way to manage your application's dependencies
 homepage: https://bundler.io
--- a/.licenses/bundler/faraday.dep.yml
+++ b/.licenses/bundler/faraday.dep.yml
@@ -1,6 +1,6 @@
 ---
 name: faraday
-version: 1.9.3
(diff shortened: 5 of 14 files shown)


jonabc:
The `licensed-ci` GitHub Action has created a [pull request](https://github.com/github/licensed/pull/456) containing license metadata updates based on the changes in this branch.

villelahdenvuo (approved):
LGTM 👍

jonabc:
@benbalter is there an ETA on the next licensee release?  I could probably get by on merging and releasing this change without https://github.com/licensee/licensee/pull/533, but I'd prefer to wait if a new release could happen in the near term

jonabc:
@benbalter :wave: ping on the licensee release.  Do you have time to put out a new release this week?

benbalter:
> Do you have time to put out a new release this week?

Done. https://github.com/licensee/licensee/releases/tag/v9.15.2

jonabc:
The `licensed-ci` GitHub Action has created a [pull request](https://github.com/github/licensed/pull/464) containing license metadata updates based on the changes in this branch.

docs/configuration/customizing_licensee.md
```diff
@@ -0,0 +1,13 @@
+# Customize Licensee's behavior
+
+Licensed uses [Licensee](https://github.com/licensee/licensee) to detect and evaluate OSS licenses for project dependencies found during source enumeration.  Licensed can optionally [customize Licensee's behavior](https://github.com/licensee/licensee/blob/jonabc-patch-1/docs/customizing.md#customizing-licensees-behavior) based on options set in the configuration file.
```
villelahdenvuo, line 3: I just noticed you link to your own patch branch, probably should be linked to the main repo?

villelahdenvuo:
After running `licensed env` it looks like the licensee config is ignored:

```yml
---
git_repo: true
apps:
- name: ecom-api
  source_path: "/home/runner/work/ecom-api/ecom-api"
  cache_path: "/home/runner/work/ecom-api/ecom-api/.licenses"
  sources:
  - name: ecom-api.npm
  allowed:
  - 0bsd
  - apache-2.0
  - bsd-2-clause
  - bsd-3-clause
  - cc0-1.0
  - isc
  - mit
  - mpl-2.0
  - unlicense
  - wtfpl
  ignored:
    npm:
    - "@grano/**"
    - "@fortawesome/**"
    - font-awesome
    - "@devexpress/**"
    - devexpress*
    - "@devextreme/**"
    - devextreme*
    - ap
    - map-values
    - jsonify
    - fsevents
    - lodash*
  reviewed:
    npm:
    - "@elastic/elasticsearch"
    - amazon-cognito-identity-js
    - argparse
    - color-convert
  version_strategy: git
  root: "/home/runner/work/ecom-api/ecom-api/grano-best-practices"
```

jonabc:
:wave: @villelahdenvuo the `licensed env` output is not necessarily accurate, it looks like it's hardcoded to include the information you see ☝️ .  I'll take a look at whats going on ASAP

villelahdenvuo:
@jonabc Did you have a chance to look at this? Also I think you missed my earlier comment about the doc links being broken.

jonabc:
Sorry no I haven't gotten a chance yet.  I'm going to try to look at it today or tomorrow but I can't promise anything, I've been very busy this week.  Is this blocking you from being able to use licensed?

docs/configuration/customizing_licensee.md
jonabc: ah yeah good call

villelahdenvuo:
No worries, in the end I added all the exceptions in the config, so I can use it, but it would be nice to remove as many of the manually checked licenses as possible. If you want I can open a new issue about this so you have it as a reminder.

jonabc:
@villelahdenvuo I took a quick look - have you recached licenses since updating the license config with the licensee threshold?  License determination is only done when caching dependencies, where `licensed status` is primarily reading/reporting on the cached files that are available.

I tested `crypto-js@4.1.1` in a repro project.  With the confidence threshold value set to 85 it was able to set the "mit" license, vs "other" without the threshold set.

One note - I found a bug while testing this where using the `licensed cache --force` option will not set the new license classification when the license text is unchanged. So if you changed the confidence threshold and a different license was detected the new license wouldn't be written to the cached file because the existing cached license text wouldn't have changed.  The short term workaround is to delete any files that you need updated and let licensed regenerate the file with the new classification 👍

villelahdenvuo:
That makes sense, I'll delete the cache records and try again.

villelahdenvuo:
@jonabc Do you think it would make sense to calculate a hash of the licensee config and add it as a part of the license cache records to make the cache invalidation automatic? If so, I can make a new issue about that.

jonabc:
@villelahdenvuo 🤔 I'm not certain.  It's definitely useful when the classification would change from `other` to non-`other`, but there's still a chance that invalidating based on licensee config changes could cause a manually classified non-`other` to be changed to `other`.

I'm also not sure if you are suggesting that licensed should invalidate the cache on other operations like `licensed status`.  I think in those cases I'd still prefer to avoid updating cached files during a status check.

please do open an issue though!  this is an interesting problem.  In the meantime I'm going to try to fix the direct issue where `--force` isn't working as intended.

villelahdenvuo:
@jonabc I see, I think the use of "cache" here is a bit misleading, because cache is usually considered something that is meant to speed up processing, but I guess with licensed people can use it to manually review licenses, so it's more like a license database.

I thought that the reviewing was meant to be done in the licensed configuration file and the cached records shouldn't be touched.

I'm not sure what kind of issue I should make for this. Maybe it's a documentation issue after all.

jonabc:
> I'm not sure what kind of issue I should make for this. Maybe it's a documentation issue after all.

Yeah I'm not too sure either 😅 .  I guess the question I'd ask is "what could be improved?", and if anything comes to mind then please do open an issue.  Terminology and words in general can be a hard problem.

villelahdenvuo:
@jonabc I opened #475 to address the documentation issue. I think for me the biggest issue in the beginning was that there was no "getting started" section, only list of commands I had to guess how they work together to build a workflow (licensed-ci helped there), but it still didn't clarify what's the expected/default workflow.

PR #480, opened 2022-03-15
Use pip list to get packages for pip source

jonabc:
This change updates the pip source to gather dependencies from `pip list` rather than parsing `requirements.txt`.  This is done to accommodate the multitude of [requirement specifies](https://pip.pypa.io/en/stable/cli/pip_install/#requirement-specifiers) that weren't currently being handled, such as direct file specifiers.  Instead licensed now relies on pip to report what dependencies are being used.  As a side effect, this now causes the pip source to report on transitive/nested dependencies, where previously only explicit dependencies were being reported 🤦 

After all dependencies were being reported, I found that performance was really suffering because many more `pip show` commands were being executed. Thankfully though `pip show` supports entering multiple package names in one command, which overall made the source faster than it was before this PR 🚀 .

Some minor side effects of this change
1. the number of test cases needed drops significantly because licensed no longer needs to validate requirements.txt parsing logic
1. there is no longer a need for a `requirements.txt` file.  we just need to know the virtual env directory to run under and pip gives us all the rest of the information.

### DIFF
--- a/.licenses/bundler/bundler.dep.yml
+++ b/.licenses/bundler/bundler.dep.yml
@@ -1,6 +1,6 @@
 ---
 name: bundler
-version: 2.3.8
+version: 2.3.9
 type: bundler
 summary: The best way to manage your application's dependencies
 homepage: https://bundler.io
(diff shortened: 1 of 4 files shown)


jonabc:
The `licensed-ci` GitHub Action has created a [pull request](https://github.com/github/licensed/pull/481) containing license metadata updates based on the changes in this branch.

PR #486, opened 2022-03-17
Consider Composer dev dependencies

digilist:
I noticed that composer dev dependencies haven't been considered by this package, so I wanted to add this.

As I am not sure about the BC policy of this library, I made this a configuration option, so that people do not automatically get all the license updates. In general I would consider this enabled by default though, wdyt?

### DIFF
--- a/docs/sources/composer.md
+++ b/docs/sources/composer.md
@@ -12,3 +12,12 @@ The default composer application file location is `<repository root>/composer.ph
 composer:
   application_path: "/path/to/composer"
 ```
+
+### Dev dependencies
+
+By default licensed ignores all dev dependencies. To consider dev dependencies as well, use the `composer.include_dev` configuration setting.
+
+```yml
+composer:
+  include_dev: true
+```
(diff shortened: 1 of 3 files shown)


jonabc (commented):
🚀 ✨  thanks for opening this PR!  One small suggestion to keep ruby style consistency, and can you update the composer fixtures and add a test case that exercises the new config option?  Let me know if any of how to make that work is unclear!

> As I am not sure about the BC policy of this library, I made this a configuration option, so that people do not automatically get all the license updates. In general I would consider this enabled by default though, wdyt?

Dev or test dependencies do not normally ship with a product, so are not normally a primary concern where OSS compliance is involved.  I think leaving the default to false and making this opt-in like the solution in this PR is the best strategy 👍

digilist:
Thanks for the quick feedback! I guess I was just happy to get the code working and forgot about testing :sweat_smile:

jonabc (approved):
🚀  looks great, thanks!  I'm putting out a release for licensed later today so this should be available for use relatively soon 👍  

I'm going to ignore the license metadata verification failure, the issue is related to the running the workflow and not anything specific to the code changed here.

digilist:
Thank you very much! :+1:

PR #504, opened 2022-04-25
Support pip packages that install with a license_files folder

jonabc:
From https://peps.python.org/pep-0639/ it looks like pip can start to install dependencies where the dependency license metadata files are installed at `.dist-info/license_files`.  This PR accounts for that change by having the source enumerator check whether the `license_files` folder exists under .dist-info, and using that folder to look for license files if it does exist.

I've also changed the pipenv source to inherit from the pip source.  These two sources started with different implementations but over time they have gotten pretty close to each other, where now they are only slightly different.  This change consolidates the implementation so that both sources benefit from future changes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -437,7 +437,7 @@ jobs:
     needs: core
     strategy:
       matrix:
-        python: [ '3.6', '3.7', '3.8', '3.9' ]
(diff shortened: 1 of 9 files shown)


jonabc:
The `licensed-ci` GitHub Action has created a [pull request](https://github.com/github/licensed/pull/505) containing license metadata updates based on the changes in this branch.

PR #524, opened 2022-07-28
Ignore swift build artifacts

CvX:


jonabc (approved):

jonabc:
The CI failures are unrelated to the change and are most likely due to not pinning to specific version of rubocop related gems.  I'm going to fix those but you'll need to merge the changes down into your branch afterwards in order to get CI working again here and on your other fixes.  Sorry for the trouble!

jonabc:
Looks good, I'm ignoring the non-blocking cached license failure workflow, which is due to the token not being available on PRs from forks.

PR #525, opened 2022-07-28
Fix cargo setup

CvX:
Previously `script/setup -f` would delete all cargo test files

### DIFF
--- a/script/source-setup/cargo
+++ b/script/source-setup/cargo
@@ -13,7 +13,11 @@ BASE_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 cd $BASE_PATH/test/fixtures/cargo
 
 if [ "$1" == "-f" ]; then
-  find . -not -regex "\.*" -and -not -path "*app*" -print0 | xargs -0 rm -rf
+  find . -not -regex "\.*" \
+    -and -not -path "*/Cargo.lock" \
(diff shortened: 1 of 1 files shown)


jonabc:
:wave: Thanks for tackling this!  The changes don't look like an issue, but would you happen to know if the issue this is fixing is due to changed behavior in newer versions of cargo?  The setup script and the test fixtures used to work 😅

CvX:
`script/setup -f` deleting cargo fixture files I believe was always the issue. It was removing all files but those with `*app*` in their name - and there were no such files 😅 I think it was just a copy-and-paste mistake, the line was most likely borrowed from https://github.com/github/licensed/blob/master/script/source-setup/cabal#L16-L16

The compiling issue, turns out, not an issue! 😃 I was fooled by an error message I got when still trying to fix the setup script. I now removed that part of the change.

jonabc:
> script/setup -f deleting cargo fixture files I believe was always an issue. It was removing all files but those with *app* in their name - and there were no such files 😅 I think it was just a copy-and-paste mistake, the line was most likely borrowed from https://github.com/github/licensed/blob/master/script/source-setup/cabal#L16-L16

🤦 oh, yup.  Thanks for catching and fixing 🙇

jonabc (approved):

jonabc:
Looks good, I'm ignoring the non-blocking cached license failure workflow, which is due to the token not being available on PRs from forks.

PR #528, opened 2022-07-30
Replace mini_racer to fix latest macOS/Apple silicon compatibility

CvX:
`mini_racer` 0.3.1 isn't compatible with more recent macOS versions (in part due to lack of python 2) which made `script/setup` fail with `Gem::Ext::BuildError: ERROR: Failed to build gem native extension`

Updating `mini_racer` caused a different problem - a bug in bundler caused it to install both `linux` and `linux-musl` variants in the CI environment. (see: https://github.com/rubyjs/libv8-node/issues/5 and https://github.com/rubygems/rubygems/issues/3174)

This instead replaces `mini_racer` example with `loofah`. That gem depends on `nokogiri` which provides different binary variants, like `libv8`/`libv8-node` did.

But this one builds properly on macOS and does not provide `linux-musl` variant so it should not cause any issues on CI.

### DIFF
--- a/test/fixtures/bundler/Gemfile
+++ b/test/fixtures/bundler/Gemfile
@@ -26,8 +26,7 @@ end
 gem "pathed-gem-fixture", path: "pathed-gem-fixture"
 
 # verify https://github.com/github/licensed/issues/71
-git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }
-gem "mini_racer", "0.3.1"
+gem "loofah", "2.18.0"
 
 # verify https://github.com/github/licensed/issues/153
 gem "aws-sdk-core", "3.39.0"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,7 +38,7 @@ jobs:
     needs: core
     strategy:
       matrix:
-        bundler: [ '~> 1.17.0', '~> 2.0.0', '~> 2.1.0', '~> 2.2.0', '~>2.3.0' ]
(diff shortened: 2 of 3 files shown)


test/fixtures/bundler/Gemfile
```diff
@@ -26,8 +26,7 @@ end
 gem "pathed-gem-fixture", path: "pathed-gem-fixture"
 
 # verify https://github.com/github/licensed/issues/71
-git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }
-gem "mini_racer", "0.3.1"
+gem "loofah", "2.18.0"
```
jonabc, line 29: can this use nokogiri directly or do you know if having it as an indirect dependency makes a difference?  it's been awhile since that issue was opened and then fixed, but I think mini_racer was used here instead of libv8 because it matched the repro steps of the issue.

if we need to break away from libv8, maybe we can also get away from an indirect dependency?

jonabc: oh also, can you add a comment so that it's clear whats being tested since the repro doesn't match the issue link any longer?  feel free to ignore my suggested text and use your own if you would prefer 👍 

```suggestion
# depends on platform-dependent nokogiri installation
gem "loofah", "2.18.0"
```

jonabc (approved):
✨ The changes look great and I appreciate you working around the problem.  left a few comments that I wanted to get your thoughts on before approving but nothing is blocking.

jonabc:
I'm going to merge these changes as-is to fix the build on the bundler test fixtures.  It looks like something changed in the most recent bundler release that is causing problems with this tool and I need to be able to run the tests to figure out the fix 😄 .  I might update the comments and such later.

Issue #529, opened 2022-07-30
Issue with Ruby v3, mac and yarn project


tom-quiltt:
When I use `github/licensed` in Ruby v3 environment, it works fine for bundler project but error out when I run `licensed status` or `licensed list` for yarn project.

When I switch to Ruby v2.7, it works fine.

Note: It might be mac specific issue, Ruby v3 with licensed works fine on linux when running yarn project.

```sh
> ruby -v
ruby 3.0.4p208 (2022-04-12 revision 3fa771dded) [arm64-darwin21]
```
```sh
  yarn
RSTRING_PTR is returning NULL!! SIGSEGV is highly expected to follow immediately. If you could reproduce, attach your debugger here, and look at the passed string.
/Users/tomlee/.rbenv/versions/3.0.4/lib/ruby/gems/3.0.0/gems/json-2.6.1/lib/json/common.rb:216: [BUG] Segmentation fault at 0x0000000000000000
ruby 3.0.4p208 (2022-04-12 revision 3fa771dded) [arm64-darwin21]

-- Crash Report log information --------------------------------------------
   See Crash Report log file under the one of following:                    
     * ~/Library/Logs/DiagnosticReports                                     
     * /Library/Logs/DiagnosticReports                                      
   for more details.                                                        
Don't forget to include the above Crash Report log file in bug reports.     

-- Control frame information -----------------------------------------------
c:0034 p:---- s:0197 e:000196 CFUNC  :parse
```

jonabc:
thanks for opening this.

It looks like the [bug you're hitting](https://github.com/flori/json/issues/495) is with the json gem in version 2.6.1 which [was fixed](https://github.com/flori/json/commit/b59368a8c23976d9e44adc8f8c284fdd954a0d33) and included in the 2.6.2 release.

I was previously not setting any explicit requirement on the json gem in the gemspec because it's shipped with ruby. To fix this for versions of ruby that ship with the affected version of the json code, I'll need to make the requirement explicit.  Should be a quick fix 👍

tom-quiltt:
Thank you so much for fixing this!

PR #530, opened 2022-08-01
Add explicit dependency on JSON gem

jonabc:
closes https://github.com/github/licensed/issues/529

Add an explicit dependency on the json gem >= 2.6.2 to include a [fix](https://github.com/flori/json/commit/b59368a8c23976d9e44adc8f8c284fdd954a0d33) on a [segfault in json@2.6.1](https://github.com/flori/json/issues/495) with ruby 3.0.4.

I opted to not put any sort of upper bound on the dependency range to match the pre-existing behavior as much as possible.  Before this change whatever version of the json gem was included with a ruby installation would be the version of the gem that gets used.  That versioning scheme has no upper bound, so I excluded it in this change

### DIFF
--- a/.licensed.yml
+++ b/.licensed.yml
@@ -7,6 +7,7 @@ allowed:
 
 reviewed:
   bundler:
+    - json
     - pathname-common_prefix
     - racc
     - reverse_markdown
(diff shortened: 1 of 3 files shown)


jonabc:
The `licensed-ci` GitHub Action has created a [pull request](https://github.com/github/licensed/pull/531) containing license metadata updates based on the changes in this branch.

PR #535, opened 2022-08-17
Handle changes in materializing lazy specs in recent versions of bundler

jonabc:
Seen in [CI logs on a dependency update](https://github.com/github/licensed/runs/7841329149?check_suite_focus=true), licensed isn't working as expected for recent versions of bundler.  As far as I can tell this has to do with small changes in the materialization of lazy specifications, and theres a new method `materialize_for_installation` that I can tie into to check whether materialization failed.  From what I can tell it should play fine with the previous `__materialize__` override and should only affect newer versions of bundler that implement the method on LazySpecifications.

### DIFF
--- a/.licenses/bundler/bundler.dep.yml
+++ b/.licenses/bundler/bundler.dep.yml
@@ -1,6 +1,6 @@
 ---
 name: bundler
-version: 2.3.19
+version: 2.3.20
 type: bundler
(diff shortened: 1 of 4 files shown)


jonabc:
The `licensed-ci` GitHub Action has created a [pull request](https://github.com/github/licensed/pull/536) containing license metadata updates based on the changes in this branch.

PR #538, opened 2022-08-19
Include #licensed-support reference

tyliec:
Add name of slack channel


jonabc (changes requested):
This is a public repo and shouldn't have any GitHub specific details :)

PR #547, opened 2022-10-06
Find license files for hatch-built python dependencies

jonabc:
Python wheels built using hatch installs files in [a "licenses" folder](https://github.com/pypa/hatch/blob/b67a49aa6384269a027a91f845fcacbac3f836ee/backend/src/hatchling/builders/wheel.py#L511) under `dist-info`, which was not previously being looked at.  Longer term it might be better to examine all files under `dist-info` for license contents but this is an ok option in the short term to fix the immediate issue where license files aren't found.


jonabc:
the licenses CI job didn't create a PR to update the out of date dependency metadata because this branch name ends in `-licenses` 🤦 .  going to merge this and handle the metadata update when the check runs against the default branch

ofek:
fyi that directory will be the new standard https://peps.python.org/pep-0639/

PR #553, opened 2022-10-14
Try fixing rubocop CI failures

jonabc:
Looks like maybe the new release from rubocop-github is causing some issues here.  I'm trying to fix it by removing the explicit dependency version range from licensed's gemspec, which should pick up newer versions of rubocop and let licensed be a bit more flexible in usage.


jonabc:
The `licensed-ci` GitHub Action has created a [pull request](https://github.com/github/licensed/pull/554) containing license metadata updates based on the changes in this branch.

Issue #541, opened 2022-09-27
Support specifying version (ranges) on reviewed dependencies


ericcornelissen:
Related to some other issues (#196, #418, #442) but I believe slightly different: my request is to allow for specifying versions (or version ranges) on reviewed dependencies such that `licensed status` errors if the version doesn't match.

I'm aware of the ["license needs review" state](https://github.com/github/licensed/blob/master/docs/commands/status.md#license-text-has-changed-and-needs-re-review-if-the-new-text-is-ok-remove-the-review_changed_license-flag-from-the-cached-record) (originating from #244). However, this requires that the Licensed cache is committed to VCS (which I'm trying to avoid because I believe it doesn't fit my use case). If I can specify a version on reviewed dependency, Licensed knows which version I reviewed so can tell me to re-review if the version changes.

Supporting version ranges as well could be seen as a quality of life feature, particularly given that I've seen it argued that license changes should be considered a breaking change (unfortunately I don't have a reference for that...).

I have two (non-mutually exclusive) suggestions for how to support this from the configuration side:

1. Add support for adding `@x.y.z` to the package name. If present, the version is compared against the actual version of the dependency. If not present, Licensed's behavior is unchanged.
2. Update the `reviewed` configuration format to something along the lines of:

   ```yaml
   reviewed:
      npm:
        # for example
        - package: glob
          version: x.y.z

        # Or no version, same behavior as current
        - package: glob

        # Or, for backwards compatibility
        - glob
   ```

jonabc:
@ericcornelissen I'm not sure that these changes alone are going to produce the result you're looking for.  Licensed has checks which that enforce that a license file exists for each dependency.

> However, this requires that the Licensed cache is committed to VCS (which I'm trying to avoid because I believe it doesn't fit my use case).

Committing the license files to VCS is somewhat core to the purpose of licensed.  Would you mind describing your use case and maybe I can help with how to better fit licensed to what you're looking to accomplish?

ericcornelissen:
> I'm not sure that these changes alone are going to produce the result you're looking for. Licensed has checks which enforce that a license file exists for each dependency.

Can you elaborate on why it won't give me the results I'm looking for. 

If I have reviewed a dependency at a particular version, I would know it doesn't have a license file. If I need to re-review the dependency because I'm now depending on a new version, I would find out it doesn't have a license file anymore.

[...]

PR #560, opened 2022-10-21
Check status of dependencies using computed metadata and configuration file

jonabc:
closes https://github.com/github/licensed/issues/541

This PR proposes a new way of running `licensed status` that does not require the prior use of `licensed cache`.

A new `--data-source` CLI option has been added to the `licensed status` CLI command, with available values of `files` and `configuration`.  `files` is the default value and maps to licensed's current behavior that requires `licensed cache` to be run and metadata files to be present in the repository under the configured `cache_path`.

Using the `configuration` option value allows small teams and projects to check their compliance without needing to first cache files.  The license metadata and contents/text are computed when the CLI command is run rather than loading data from a file.

One explicit behavior change in this PR is that dependencies added to the `reviewed` configuration must include a version identifier.  For example if licensed finds a dependency named `my_dependency` at version `v1.2.3`, the reviewed list must contain an entry for `my_dependency@v1.2.3` and not `my_dependency`.

I've tried to make the error output more helpful by including possible near-matches in common scenarios:
- when the reviewed list contains `my_dependency` rather than `my_dependency@v1.2.3`
- when the reviewed list matches a dependency at a different version

cc @ericcornelissen I'd appreciate if you can take a look and let me know if this would solve the underlying need from the linked issue you opened.  also if you wouldn't mind reading [the updated docs page](https://github.com/github/licensed/blob/reviewed-dependency-version/docs/commands/status.md) and letting me know if you think it's clear how to use the new behavior?

### DIFF
--- a/docs/commands/status.md
+++ b/docs/commands/status.md
@@ -26,8 +26,8 @@ When using `licensed status` in this scenario, licensed will compile all depende
 
 A dependency will fail the status checks if:
 
-1. The cached record's `licenses` data is empty
-2. The cached record's `license` metadata doesn't match an `allowed` license from the dependency's application configuration and the dependency has not been marked `reviewed` or `ignored`
+1. The record's `licenses` data is empty
+2. The record's `license` metadata doesn't match an `allowed` license from the dependency's application configuration and the dependency has not been marked `reviewed` or `ignored`
    - If `license: other` is specified and all of the `licenses` entries match an `allowed` license a failure will not be logged
    - A `reviewed` entry must reference a specific version of the depdency, e.g. `<name>@<version>`.  The version identifier must specify a specific dependency version, ranges are not allowed.
 
--- a/docs/commands/status.md
+++ b/docs/commands/status.md
@@ -1,16 +1,35 @@
 # `licensed status`
 
-The status command finds all dependencies and checks whether each dependency has a valid cached record.
+The status command finds all dependencies and checks whether each dependency has a valid record.  There are two methods for checking dependencies' statuses
+
+## Checking status with metadata loaded from cached files
+
+This is the default method for checking the status for dependencies and the recommended method for using licensed.  Checking status from cached metadata files will occur when the `--data-source` CLI flag is either unset, or set to `--data-source=files`.
+
+When using `licensed status` in this scenario, licensed will only compile a minimal amount of dependency metadata like the dependency name and version to match against cached files.  Dependency license and notice texts are loaded from cached files that are created from the [licensed cache](./cache.md) command.
 
 A dependency will fail the status checks if:
 
 1. No cached record is found
 2. The cached record's version is different than the current dependency's version
 3. The cached record's `licenses` data is empty
-4. The cached record's `license` metadata doesn't match an `allowed` license from the dependency's application configuration.
+4. The cached record's `license` metadata doesn't match an `allowed` license from the dependency's application configuration and the dependency has not been marked `reviewed` or `ignored`
    - If `license: other` is specified and all of the `licenses` entries match an `allowed` license a failure will not be logged
 5. The cached record is flagged for re-review.
    - This occurs when the record's license text has changed since the record was reviewed.
+
+## Checking status with computed metadata
+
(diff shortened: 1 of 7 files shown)


docs/commands/status.md
```diff
@@ -1,17 +1,36 @@
...
+## Checking status with computed metadata
+
+Checking status with computed metadata and the licensed configuration will occur when the `--data-source` CLI flag is set to `--data-source=configuration`.
+
+When using `licensed status` in this scenario, licensed will compile all dependency metadata and license text to use in status checks.  There is no need to run `licensed cache` prior to running `licensed status --data-source=configuration`.
+
+A dependency will fail the status checks if:
+
+1. The cached record's `licenses` data is empty
+2. The cached record's `license` metadata doesn't match an `allowed` license from the dependency's application configuration and the dependency has not been marked `reviewed` or `ignored`
```
ericcornelissen, line 30: I feel like using "cached record" here might be confusing

```suggestion
1. The record's `licenses` data is empty
2. The record's `license` metadata doesn't match an `allowed` license from the dependency's application configuration and the dependency has not been marked `reviewed` or `ignored`
```

Maybe further rewording this to use the term "dependency" instead of "record" would be even better, but for me "record" suffices.

docs/commands/status.md
```diff
@@ -20,6 +39,9 @@ A dependency will fail the status checks if:
    - default value: not set, all configured sources
 - `--format`/`-f`: the output format
    - default value: `yaml`
+- `--data-source`/`-d`: where to find the data source of records for status checks
+   - available values: `files`, `configuration`
+   - default value: `files`
```
ericcornelissen, line 44: Just a suggestion, I feel like renaming the available value `files` to `cache` would be clearer.

ericcornelissen (commented):
> I'd appreciate if you can take a look and let me know if this would solve the underlying need from the linked issue you opened.

I think this makes a lot of sense and meets the underlying needs from my perspective.

docs/commands/status.md
jonabc: 😆 that's what I started out calling that option, but I ended up at files because I've gotten feedback in the past that the term `cache` is confusing related to storing files in a repo.  I'm pretty ambivalent myself so if you feel at all strongly about naming this `cache` I'm happy to rename.

docs/commands/status.md
jonabc: I'm going to leave this as `record`, but 👍  to removing `cached`  The dependency is computed live on every licensed execution while the record may not be, especially in the case of using files written to a repo.

As an example, the record may have been written to reference a dependency at version 1.0.0, while the live computed dependency is at 2.0.0.  These types of mismatches can occur with all of the fields that cause a `licensed status` error.

jonabc:
@mlinksva curious to get your take here as well.  The crux of this change is to allow users to specify something like the following in the configuration file in lieu of writing metadata files into a repo.  

```yaml
reviewed:
   bundler:
      - licensed@1.2.3
```

When using the new CLI option all of the existing checks on license text and other metadata still apply except for the existence of the metadata file.  The checks are just run against a live computation of the dependency metadata instead metadata read from a file in the repo.  When checking status using the new behaviors, a `reviewed` dependency entry *must* include a version identifier to avoid something like `licensed@*` or `licensed@>= 0.0` which I imagine is not great for compliance purposes 😆 .

docs/commands/status.md
ericcornelissen: I don't feel strongly about it either, I just thought that it aligns better with terminology used elsewhere in the project, e.g. the `licensed cache` command.

mlinksva:
Requiring a version for dependencies in `reviewed` does seem like a good safeguard` as if the license texts aren't cached you wouldn't see any changes if they had changed, which they could between versions, thus meriting updating review for a new version. If this is the idea 👍

ericcornelissen:
@jonabc, do you think it would be possible to add support for using computed metadata to the [`licensed notice`](https://github.com/github/licensed/blob/77e05ec1268889b49da479010adb9f33c784968c/docs/commands/notices.md) command as well?

jonabc:
@ericcornelissen sure https://github.com/github/licensed/pull/572

PR #572, opened 2022-11-12
Allow generating notice file from live computed data

jonabc:
addresses https://github.com/github/licensed/pull/560#issuecomment-1312449501

This adds a `--computed` long form, `-l` short form option to `licensed notice`.  When set, the command will generate a NOTICE file from dependency data computed dynamically, rather than loading dependency data from metadata files.

### DIFF
--- a/lib/licensed/cli.rb
+++ b/lib/licensed/cli.rb
@@ -46,13 +46,15 @@ module Licensed
       run Licensed::Commands::List.new(config: config), sources: options[:sources], reporter: options[:format], licenses: options[:licenses]
     end
 
-    desc "notices", "Generate a NOTICE file from cached records"
+    desc "notices", "Generate a NOTICE file with dependency data"
     method_option :config, aliases: "-c", type: :string,
       desc: "Path to licensed configuration file"
     method_option :sources, aliases: "-s", type: :array,
       desc: "Individual source(s) to evaluate.  Must also be enabled via configuration."
+    method_option :computed, aliases: "-l", type: :boolean,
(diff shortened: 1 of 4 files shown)


jonabc:
The `licensed-ci` GitHub Action has created a [pull request](https://github.com/github/licensed/pull/573) containing license metadata updates based on the changes in this branch.

lib/licensed/cli.rb
```diff
@@ -46,13 +46,15 @@ def list
...
     end
 
-    desc "notices", "Generate a NOTICE file from cached records"
+    desc "notices", "Generate a NOTICE file with dependency data"
     method_option :config, aliases: "-c", type: :string,
       desc: "Path to licensed configuration file"
     method_option :sources, aliases: "-s", type: :array,
       desc: "Individual source(s) to evaluate.  Must also be enabled via configuration."
+    method_option :computed, aliases: "-l", type: :boolean,
+      desc: "Whether to generate a NOTICE file using computed data or cached records"
```
ericcornelissen, line 55: Should this option be added to [the `licensed notices` command documentation](https://github.com/github/licensed/blob/c103b2ff8108c05f5bc79e1eae3527dfde20293e/docs/commands/notices.md)?

ericcornelissen (commented):
This is awesome! Thanks for the quick response to my question :smile:

lib/licensed/cli.rb
jonabc: 🤦 yep, I made those changes but looks like I never pushed them up to GitHub.  Thanks for catching that!

Issue #577, opened 2022-11-17
False review_changed_license when release changed.


schmidtw:
I just ran into a falsely flagged license change review request.  The licenses and patent files are both identical in the cache except the version.

Old cached file:
```yaml
---
name: golang.org/x/sys/unix
version: v0.1.0
type: go
summary: Package unix contains an interface to the low-level operating system primitives.
homepage: https://pkg.go.dev/golang.org/x/sys/unix
license: other
licenses:
- sources: sys@v0.1.0/LICENSE
  text: |
    Copyright (c) 2009 The Go Authors. All rights reserved.

    Redistribution and use in source and binary forms, with or without
    modification, are permitted provided that the following conditions are
    met:

       * Redistributions of source code must retain the above copyright
    notice, this list of conditions and the following disclaimer.
       * Redistributions in binary form must reproduce the above
    copyright notice, this list of conditions and the following disclaimer
    in the documentation and/or other materials provided with the
    distribution.
       * Neither the name of Google Inc. nor the names of its
    contributors may be used to endorse or promote products derived from
    this software without specific prior written permission.

    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
    "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
    LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
    A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
    OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
    SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
    LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
    DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
    THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
    (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
[...]

PR #578, opened 2022-11-25
Use raw source paths when creating sort key for license information

jonabc:
closes https://github.com/github/licensed/issues/577

The underlying problem from the linked issue was that the sort key returned by `DependencyRecord::License` was taking a digest of the source paths.  That was implemented under the assumption that the path of a license file would be the same between versions, all else being equal.

When using go modules though, that is an invalid assumption 😅 .  The path can contain version information, e.g. `sys@v0.1.0/LICENSE` and `sys@v0.1.0/PATENTS` vs `sys@v0.2.0/LICENSE` and `sys@v0.2.0/PATENTS`.  When the paths were included as part of the hash digest, this resulted in the equivalent files being ordered differently:
- for `v0.1.0`, the order was `LICENSE`, `PATENTS`
- for `v0.2.0`, the order was `PATENTS`, `LICENSE`

To correct that, I'm moving the source paths out of the hash digest and using a string sort key rather than an integer.  The key is used to sort files within the same record, and using the string sort should accomplish the desired path comparison behavior.

The downside to this change is that sorting the contents will take a little more time.  The impact should be small though, most dependencies have a very small number of license files.


jonabc:
The `licensed-ci` GitHub Action has created a [pull request](https://github.com/github/licensed/pull/579) containing license metadata updates based on the changes in this branch.

jonabc:
ok, after some hygiene work to get go tests working again in CI (as a separate PR) this is looking good.  merging

PR #580, opened 2022-11-25
bump versions of go used in tests

jonabc:
testing was done on versions 1.12 -> 1.16... all of which are old and started to cause issues in CI.  It looks like 1.18 and 1.19 are the versions currently being actively supported, and I've added 1.17 in case not everyone has updated at this point.  It's not feasible to test against every version of go but hopefully one unsupported prior version won't be a big deal 🤷


jonabc:
The `licensed-ci` GitHub Action has created a [pull request](https://github.com/github/licensed/pull/581) containing license metadata updates based on the changes in this branch.

jonabc:
the drift between go with modules and go without modules was getting difficult to manage so I ended up removing testing support for go without modules.  To do that I moved the existing go without modules fixtures to a separate `test/fixtures/dep` folder, with a new `script/source-setup/dep` script and corresponding changes needed in various fixtures and test scripting

PR #470, opened 2022-03-01
Support gradle multi-project

civitaspo:
Signed-off-by: civitaspo <civitaspo@gmail.com>

Fix https://github.com/github/licensed/issues/301

This PR supports gradle projects with multi-project configuration. The `settings.gradle` defined in the project was ignored when executing `printDependencies`, which is a task to output dependencies, so it is now used. Running a Gradle Task without `settings.gradle` in a project that defines `settings.gradle` could lead to unintended behavior, so this fix ensures that Gradle Tasks run correctly. Just to make sure that backwards compatibility is not broken, I've tested this on a few projects I manage and found no change in behavior. The support for `build.gradle.kts` and `settings.gradle.kts` written in Kotlin mentioned in https://github.com/github/licensed/issues/301 will be done in the next PR after this PR is merged.

Issue #301, opened 2020-08-27
Add support for Gradle multi-module projects.


CristianGM:
The approach to inject the plugin dependency-license-report plugin used [here](https://github.com/github/licensed/blob/master/lib/licensed/sources/gradle.rb#L147) is making a modified copy of the build.gradle and running it with `-b`, it means Gradle will ignore the settings.gradle and ignore all the subprojects (modules).

I asked on Gradle Slack and they told me to use an init script instead so it works properly, that will also fix the issue that we are only considering `build.gradle` and not the "new" `build.gradle.kts` which would change the syntax slightly.

Alternatively it would be possible to copy the settings.gradle to another file adding to it `rootProject.buildFileName = "license-....gradle"`´ and using it adding to the cli `-c licensed-settings.gradle`, but then we still need to add support for `.gradle.kts` files

jonabc:
@CristianGM I don't have deep knowledge of Gradle, but am happy with any changes that maintain backwards compatibility while solving the problems you mention. Would you be able to open a PR?

jonabc:
This was resolved with https://github.com/github/licensed/pull/583, which will be part of the major version 4.0 release

PR #583, opened 2022-12-02
Support Gradle multi-project builds

LouisBoudreau:
## Description

The goal of this PR is to add support for Gradle [multi project builds](https://docs.gradle.org/current/userguide/multi_project_builds.html). To do so I implemented the suggestion provided in this [issue](https://github.com/github/licensed/issues/301) that recommended the use of an init script. 

The approach used is different from the other opened [PR](https://github.com/github/licensed/pull/470#issue-1155622261) to fix this issue.

I also upgraded to the latest version of the [gradle plugin](https://plugins.gradle.org/plugin/com.github.jk1.dependency-license-report).

## Tests

I have split the gradle fixtures, one single project and one multi project. I also added some tests for the multi-project implementation.


lib/licensed/sources/gradle.rb
```diff
@@ -84,16 +91,23 @@ def enabled?
...
+          project: project
+        )).map do |package|
+          name = "#{package['group']}:#{package['name']}"
           Dependency.new(
             name: name,
             version: package["version"],
             path: config.pwd,
             executable: gradle_executable,
             configurations: configurations,
+            project: project,
```
jonabc, line 108: is it possible for different projects in a multi-project build to have different versions of the same dependency? 

If yes, this value should be part of the `name:` argument rather than being added as a new property value.

lib/licensed/sources/gradle.rb
```diff
@@ -122,49 +138,51 @@ def configurations
...
-      def self.add_gradle_license_report_plugins_block(gradle_build_file)
-
-        if gradle_build_file.include? "plugins"
-          gradle_build_file.gsub(/(?<=plugins)\s+{/, " { id 'com.github.jk1.dependency-license-report' version '1.16'")
-        else
-
-          gradle_build_file = " plugins { id 'com.github.jk1.dependency-license-report' version '1.16' }" + gradle_build_file
-        end
+      def project
+        @project ||= config.dig("gradle", "project")
```
jonabc, line 142: A few questions about this new value:

- Is a default value needed?  What's the expected behavior if a user doesn't specify `project` in their configuration?  
- Is there support for a user to run licensed over every project in a multi-project build?  If so, how does a user specify that?

lib/licensed/sources/gradle.rb
```diff
@@ -122,49 +138,51 @@ def configurations
...
       end
 
-      def self.gradle_command(*args, path:, executable:, configurations:)
-        gradle_build_file = File.read("build.gradle")
-
-        if !gradle_build_file.include? "dependency-license-report"
-          gradle_build_file = Licensed::Sources::Gradle.add_gradle_license_report_plugins_block(gradle_build_file)
-        end
-
+      def self.gradle_command(*args, path:, executable:, configurations:, project:)
```
jonabc, line 145: it looks like you're getting rid of the `add_grade_license_report_plugins_block` function, as well as changing the name and contents of the created file.

Would you mind adding some context behind the purpose of the changes and any specifics about the implementation?  

Also, it looks like you're using some new properties in the init script, do you know how far back that new syntax is supported?  Licensed generally aims to support all LTS versions of tools/frameworks so I want to make sure this change won't cause any backwards compatibility failures for users on older supported versions of gradle.

jonabc (commented):
Thanks for opening this!  I'm excited to see the gradle source getting updated.

😓 apologies if any of my questions are simplistic, I don't use gradle at all in my day-to-day development work and am pretty unfamiliar with how it works.

Along with the comments left on the change, a few higher level comments:

- can you move `.gitignore` back to `test/fixtures/gradle`?  It looks like both single and multi project fixtures use the same gitignore, and it would be preferable to keep the file closer to root
- based on ☝️ , can you remove `test/fixtures/gradle/.gradle` from this change?  I don't believe that folder is meant to be checked in based on the `.gitignore` file, but it wasn't ignored because `.gitignore` was moved to the different fixture folders
- can you update the gradle documentation with anything that users need to know for how to use the new multi-project support, including the new `gradle.project` config value?

lib/licensed/sources/gradle.rb
```diff
@@ -122,61 +76,116 @@ def configurations
...
-                  def dependencies = []
-                  configs.each {
-                      configurations[it].resolvedConfiguration.resolvedArtifacts.each { artifact ->
-                          def id = artifact.moduleVersion.id
-                          dependencies << "  { \\"group\\": \\"${id.group}\\", \\"name\\": \\"${id.name}\\", \\"version\\": \\"${id.version}\\" }"
+            "gradle" if Licensed::Shell.tool_available?("gradle")
+          end
+        end
+
+        def init_script(configurations)
```
LouisBoudreau, line 148: > it looks like you're getting rid of the add_grade_license_report_plugins_block function, as well as changing the name and contents of the created file.
> Would you mind adding some context behind the purpose of the changes and any specifics about the implementation?

Of course!

Here are the two reasons why I think an [init script](https://docs.gradle.org/current/userguide/init_scripts.html) is a great approach.

1. [Previously](https://github.com/github/licensed/pull/583/files#diff-155ac2e2d3b0e4ce3b140ae23797b529bba82dd79608e6b2c6134b25a11d8293L131), the `plugins` block was added at the beginning of the `build.gradle`. This approach raises an error when the `build.gradle` file has a `buildScript` block. 

    ```buildscript {} blocks must appear before any plugins {} blocks```

     By using an init script we can initialize the [Gradle License Plugin](https://github.com/jk1/Gradle-License-Report) for all subprojects without being prone to errors related to order of configuration keys in `build.gradle` files.

2. Previously, the modification of the build.gradle file was only done to one file, the one present in the folder where we run `licensed`.  This meant that the plugin and the `printDependencies` task was not initialized for subprojects. Also running `licensed` directly from a subproject folder was raising an error since no `settings.gradle` file was present. The `settings.gradle` file is usually present at the root of the build projet. 

    By using an init script we can initialize the `printDependencies` in all subprojects and safely execute it from to root by[ prefixing the subproject name](https://docs.gradle.org/current/userguide/intro_multi_project_builds.html#sec:executing_tasks_by_name). Ex: `subproject:printDependencies`

lib/licensed/sources/gradle.rb
LouisBoudreau: After reflexion I thought that the `apps` configuration key was a great option for multi-project builds. 
I therefore removed the `project` key and updated the code. 

To run licensed over multiple projects who shares the same licensed configuration you could do:

```
apps:
   - source_path: path/to/subprojectA
   - source_path: path/to/subprojectB|
   

what do you think of that?

File: lib/licensed/sources/gradle.rb (hunk 3 above)
Comment (LouisBoudreau) [reply lib/licensed/sources/gradle.rb]: https://github.com/github/licensed/pull/583#discussion_r1041383560

File: .github/workflows/test.yml [hunk 5]
```diff
@@ -286,6 +286,7 @@ jobs:
         # are resolved, update this matrix list with '16'.
         # possibly fixed by https://github.com/jk1/Gradle-License-Report/pull/166
         java: [ '11' ]
+        gradle: ['current', '5.2']
```
LouisBoudreau, line 289: > Also, it looks like you're using some new properties in the init script, do you know how far back that new syntax is supported? Licensed generally aims to support all LTS versions of tools/frameworks so I want to make sure this change won't cause any backwards compatibility failures for users on older supported versions of gradle.

This is a very good point!
To simplify this task we could use the [gradle/gradle-build-action](https://github.com/gradle/gradle-build-action) to setup Gradle. 

We can then leverage the matrix to test different versions. 

This approach helps to keep track of gradle versions that are supported by licensed. 

The downside is that developers now need a local version of gradle to test  ([setup-source](https://github.com/gradle/gradle-build-action)), but this is also the case for other package managers.

jonabc: > The downside is that developers now need a local version of gradle to test ([setup-source](https://github.com/gradle/gradle-build-action)), but this is also the case for other package managers.

I think that would be a requirement for this PR anyway though, since the changes here get rid of the local `gradlew` executable.  As far as I know, the only reason there weren't previously checks around a local gradle installation was because that was included in the fixtures.  I might be wrong about that though, again I'm pretty low on contextual knowledge about gradle.

lib/licensed/sources/gradle.rb
```diff
@@ -122,61 +76,121 @@ def configurations
...
-            f.write(gradle_build_file)
-            f.write gradle_file(configurations)
-            f.close
-            Licensed::Shell.execute(executable, "-q", "-b", f.path, *args)
+        def run(*args)
+          Dir.chdir(@root_path) do
+            Tempfile.create(["init", ".gradle"], @root_path) do |f|
+              f.write(init_script(@configurations))
+              f.close
+              args << "--no-configuration-cache" if gradle_version >= "6.6"
```
LouisBoudreau, line 126: The [configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html) is an incubating feature that can be activated manually. The gradle plugin for licenses does not support it so we prevent it to run for gradle version supporting it.

lib/licensed/sources/gradle.rb
jonabc: 👍 if that works, that's a great fit with the intended use case of `apps`.  If there's a well-known pattern for naming subprojects you can also use glob ptterns like 

```yaml
apps:
  - source_path: path/to/subproject*
  
# or even forego the `apps` key and define the source_path at the root using a glob pattern
source_path: path/to/subproject*
```

script/source-setup/gradle
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+
+if [ -z "$(which gradle)" ]; then
```
jonabc, line 4: Oh hmmm if this isn't setting anything up I'm not sure that there's any reason to have this script 🤔 .  I think I misunderstood your previous comment about whether you should create this script.

Maybe it can install a gradlew executable to the `test/fixtures/gradle` subfolders?

I don't know, I guess that can be re-evaluated as a followup.

jonabc (commented):
Just to call out - I appreciate that moving code around so that the logic to run gradle is encapsulated in the new runner class, but all the code moving around makes this change a little hard for me to review and figure out what changes are functional vs what is refactoring.

An easy review in this repo is pretty important to me as I tend to be the only maintainer for the project.  The contribution model in this repo has historically looked like
1. someone adds or updates some piece of logic, or fixing a bug
1. once everything is working, they don't drop back in until there's something else broken or could be otherwise improved

I'm ok with that model and I REALLY appreciate the community support, but it means that I really need to know the code that ships with licensed so I can support it if I get a bug report that noone else is available to fix.

I think I've gotten to the bottom of this change and better understand what it's doing, but in the future I'd appreciate if you could try to limit the scope of changes and prefer multiple smaller, targeted changes vs a single big change that adds or changes functionality as well as refactoring code for readability.

I've left a few comments, I think the only comments that I'd like to see resolved before merging this are the ones related to finding the license URLs and passing the url into the dependencies in `enumerate_dependencies`

lib/licensed/sources/gradle.rb
```diff
@@ -68,46 +29,39 @@ def exist?
...
           Dependency.new(
             name: name,
             version: package["version"],
             path: config.pwd,
-            executable: gradle_executable,
-            configurations: configurations,
             metadata: {
-              "type"     => Gradle.type
+              "type" => Gradle.type,
+              "url" => csv[csv_key(name: name, version: package["version"])]
```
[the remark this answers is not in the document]
LouisBoudreau: You are absolutely right! Thanks for the suggestion

lib/licensed/sources/gradle.rb
```diff
@@ -122,61 +76,121 @@ def configurations
...
-        gradle_build_file = File.read("build.gradle")
+      # Returns the cached url for the given dependency
+      def url_for(dependency)
+        @csv[csv_key(name: dependency.name, version: dependency.version)]
+      end
 
-        if !gradle_build_file.include? "dependency-license-report"
-          gradle_build_file = Licensed::Sources::Gradle.add_gradle_license_report_plugins_block(gradle_build_file)
+      # The Gradle::Runner class is a wrapper which provides
+      # an interface to run gradle commands with the init script initialized
```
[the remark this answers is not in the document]
LouisBoudreau: This is a great idea!

lib/licensed/sources/gradle.rb
```diff
@@ -122,61 +76,121 @@ def configurations
         end
       end
 
-      def self.add_gradle_license_report_plugins_block(gradle_build_file)
+      # Prefixes the gradle command with the project name for multi-build projects.
+      def format_command(command)
+        if config.source_path != config.pwd
```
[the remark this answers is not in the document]
LouisBoudreau: This function is still relevant, here is an example of when: 

Let's say my project has the following structure:

```
project
│   settings.gradle
│   build.gradle
│   licensed.yml   
│
└──/app
  │   build.gradle
```

And the licensed.yml file content is:
```
sources:
  gradle: true
apps:
  - source_path: ./app
 ```

Then since the config file is at the root of the project and I specified the apps source path, the condition`config.source_path != config.pwd` is true. 

The gradle command will be run in the subproject context like so `app:<command>`

Does this makes sense to you ?

File: script/source-setup/gradle (hunk 7 above)
Comment (LouisBoudreau) [reply script/source-setup/gradle]: I removed it, not sure download a version of gradlew is worth it.

**Comment by LouisBoudreau** [comment]
You are absolutely right!

While applying the changes for the init script I saw an opportunity to refactor a section of code and did so without thinking about the impact it would have on the review. I apologize, in the future I will make sure to scope the changes so they are smaller and more enjoyable to review!

On another note, I wanted to point out my admiration for the work you do for Licensed. This is a really a beautiful project with excellent quality. Licensed is a project all the more relevant now that open source is more and more widespread in our codebases.

Let me know if I can help 😄

File: lib/licensed/sources/gradle.rb (hunk 10 above)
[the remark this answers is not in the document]
Comment (jonabc) [reply lib/licensed/sources/gradle.rb:84]: > Does this makes sense to you ?

Let me resay what I think you're explaining and let me know if I got it right?

When a command for the `app` subproject is run from the `project` directory, the gradle command needs to be prefixed with the project name.  Is that correct?  So to run a command on the `app` subproject while in the `project` directory, you'd need to run `app:<command>` instead of just `command`?

If so, I think I get the purpose but the usage might not be correct based on how licensed works elsewhere in the app code.  Licensed [changes the current directory](https://github.com/github/licensed/blob/bdfad9f55009596f6660d059a1c570c5919a28d5/lib/licensed/commands/command.rb#L95) to an app's `source_path` when evaluating that app.  Also, config.pwd is a [shim](https://github.com/github/licensed/blob/bdfad9f55009596f6660d059a1c570c5919a28d5/lib/licensed/configuration.rb#L57-L59) around Pathname.pwd and doesn't return the directory containing the licensed configuration file.  Unless you're manually changing the current directory, `config.source_path == config.pwd` for the duration of enumerating and evaluating dependencies for an app.

It looks like Gradle::Runner#run changes the directory to it's `@root_path` value, which is set at the creation of `gradle_runner` to `config.pwd`.  The value of `config.pwd` at that point should be equivalent to `config.source_path` though, so it's probably not actually changing the directory.

I could definitely have missed something here or misunderstood the intentions of prefixing the command with the project name though 🤷 .

File: lib/licensed/sources/gradle.rb [hunk 11]
```diff
@@ -68,46 +34,37 @@ def exist?
...
 
       def enabled?
-        !gradle_executable.to_s.empty? && File.exist?(config.pwd.join("build.gradle"))
+        gradle_runner.enabled? && File.exist?(config.pwd.join("build.gradle"))
       end
 
       def enumerate_dependencies
-        JSON.parse(self.class.gradle_command("printDependencies", path: config.pwd, executable: gradle_executable, configurations: configurations)).map do |package|
-          name = "#{package["group"]}:#{package["name"]}"
+        csv = load_csv
```
jonabc, line 49: you can remove this call, the csv will get loaded and used from the package_url method 👍 

```suggestion
```

lib/licensed/sources/gradle.rb
```diff
@@ -122,61 +79,132 @@ def configurations
...
+
+          # parse the CSV report for dependency license urls
+          gradle_licenses_dir = File.join(config.root, GRADLE_LICENSES_PATH)
+          CSV.foreach(File.join(gradle_licenses_dir, GRADLE_LICENSES_CSV_NAME), headers: true).each_with_object({}) do |row, hsh|
+            name, _, version = row["artifact"].rpartition(":")
+            key = csv_key(name: name, version: version)
+            hsh[key] = row["moduleLicenseUrl"]
+          end
+        ensure
+          FileUtils.rm_rf(gradle_licenses_dir)
```
jonabc, line 116: I'm guessing the error being thrown in CI is due to `gradle_runner.run` raising an error before `gradle_licenses_dir` is set, sorry for not accounting for that in my suggested code.  moving `gradle_licenses_dir = File.join(...)` above `gradle_runner.run("generateLicenseReport")` should fix that

jonabc (commented):
a few more comments, this is getting closer 🚀

lib/licensed/sources/gradle.rb
```diff
@@ -122,61 +78,132 @@ def configurations
         end
       end
 
-      def self.add_gradle_license_report_plugins_block(gradle_build_file)
+      # Prefixes the gradle command with the project name for multi-build projects.
+      def format_command(command)
+        if config.source_path != config.root
```
jonabc, line 83: I'm not sure that `config.root` is the right option here either.  the [config object's root](https://github.com/github/licensed/blob/master/docs/configuration/configuration_root.md) is agnostic to the requirements or concerns of any particular sources and will be one of (in order of preference):
- the configuration file's directory, when configured as `true`
- a relative path from the configuration file, when configured as a path
- the repository root, if run from a git repository
- the directory that licensed is run from, when no other values are available

Can the gradle CLI report the project name to use here based on the current directory, or at least tell whether the current project is the root or not?

jonabc (commented):
This looks like it's just stuck up now by how to find the right project name.  Also I won't be able to run CI checks until you resolve the merge conflicts on the test file.  

Sorry to cause the conflicts, I went through the test.yml workflow file to replace manual usage of `actions/cache` with setup actions' built-in caching wherever possible.

lib/licensed/sources/gradle.rb
LouisBoudreau: Great suggestion! I found a way to find the current project with the cli.

lib/licensed/sources/gradle.rb
```diff
@@ -122,61 +76,121 @@ def configurations
...
+          @executable = executable
         end
 
-        Dir.chdir(path) do
-          Tempfile.create(["license-", ".gradle"], path) do |f|
-            f.write(gradle_build_file)
-            f.write gradle_file(configurations)
-            f.close
-            Licensed::Shell.execute(executable, "-q", "-b", f.path, *args)
+        def run(*args)
```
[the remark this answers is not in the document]
LouisBoudreau: Great suggestion, I implemented it!

lib/licensed/sources/gradle.rb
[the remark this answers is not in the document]
LouisBoudreau: You are right, I applied the necessary changes

Thanks for taking the time to explain!

jonabc (approved):
:rocket: thanks so much for working through all of the feedback on this 🙇 :shipit: !

PR #584, opened 2022-12-08
Add Cocoapods source

LouisBoudreau:
The goal of this pull request is to add support for the [CocoaPods](https://cocoapods.org/) package manager

- [x] Include [documentation](https://github.com/github/licensed/blob/31b44067a069b7601060af337d7c27571486728f/docs/sources) for the new source and update the [documented source list](https://github.com/github/licensed/blob/31b44067a069b7601060af337d7c27571486728f/README.md#sources).
- [x] Add a [setup script](https://github.com/github/licensed/blob/31b44067a069b7601060af337d7c27571486728f/script/source-setup) if needed.
- [x]  Include [tests](https://github.com/github/licensed/blob/31b44067a069b7601060af337d7c27571486728f/test/source) and [test fixtures](https://github.com/github/licensed/blob/31b44067a069b7601060af337d7c27571486728f/test/fixtures) needed to verify the source in CI.
- [x]  Add a CI job to [.github/workflows/test.yml](https://github.com/github/licensed/blob/31b44067a069b7601060af337d7c27571486728f/.github/workflows/test.yml).


lib/licensed/sources/cocoapods.rb
```diff
@@ -0,0 +1,75 @@
...
+      def lockfile
+        @lockfile ||= Pod::Lockfile.from_file(config.pwd.join("Podfile.lock"))
+      end
+
+      def podfile
+        @podfile ||= Pod::Podfile.from_file(config.pwd.join("Podfile"))
+      end
+
+      def dependency_metadata(pod)
+        metadata = JSON.parse(Licensed::Shell.execute("pod", "spec", "cat", "--regex", "^#{pod.root_name}$"))
```
LouisBoudreau, line 63: I am hesitant about this and I would like to have your opinion:

This shell command fetches all the metadata for the package which is a good thing. But on the other side, it slows down the execution a lot since we need to execute it for every dependency of the project.  

Do you think the metadata is more important than the speed ?

jonabc:
:wave: @LouisBoudreau just wanted to ACK that I've seen the PR, but my days have been incredibly busy this week.  I'll take a look at this and your other open PR over the weekend.  Sorry for the delay and thank you so much for adding cocoapods support 🙇

LouisBoudreau:
Perfect @jonabc, 

Let me know if there is anything I can do 😄

jonabc (commented):
I haven't done a deep review over the entire PR yet, but I left a few comments on the new cocoapods source class.  Anything that's a `nit` is a personal preference or a small comment that won't block this from being merged 👍 

Beyond those comments, before I take another look can you add a `test/fixtures/.gitignore` for
- `Podfile.lock`
- `Pods`
- any other files that will get generated by `script/source-setup/cocoapods`

Also, can you add a `test/fixtures/command/cocoapods.yml` file with something like
```yml
expected_dependency: Chatto
source_path: test/fixtures/cocoapods
cache_path: test/fixtures/cocoapods/.licenses
sources:
  cocoapods: true
```

lib/licensed/sources/cocoapods.rb
jonabc: oh, also was there a specific reason to use the regex option rather than just passing the pods name to the CLI?

```suggestion
        metadata = JSON.parse(Licensed::Shell.execute("pod", "spec", "cat", pod.root_name))
```

LouisBoudreau: Ok thank you for your input! I will remove the CLI command in order the speedup the process. 
I you have time and you are interested, I think it would be better to implement your suggestion in a new PR. 

> oh, also was there a specific reason to use the regex option rather than just passing the pods name to the CLI?

The `pod spec cat` command returns an error when the provided pod name matches many pods. Here is an example:

CleanShot 2022-12-12 at 10 24 25@2x

By adding `^pod-name$` the spec command will execute successfully

lib/licensed/sources/cocoapods.rb
```diff
@@ -0,0 +1,75 @@
...
+          if targets_to_validate.any?
+            targets_to_validate.map(&:dependencies).flatten
+          else
+            raise Licensed::Sources::Source::Error, "Unable to find any target in the Podfile matching the ones provided in the config."
+          end
+        end
+      end
+
+      def targets
+        config.dig("cocoapods", "targets")&.map { |t| "Pods-#{t}" }
```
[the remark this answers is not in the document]
LouisBoudreau: Yes it will always be prefixed with `Pods-`, it's part of the [nomenclature](https://github.com/CocoaPods/CocoaPods/blob/123f2d02f42075fa9f3b1bc729b342bd844efb44/NOMENCLATURE.md#aggregate-target). Here is the source code responsible for this behaviour ([code](https://github.com/CocoaPods/CocoaPods/blob/98fc505fa0896ba61c0998b32ccae1c577077cc5/lib/cocoapods/installer/analyzer.rb#L382))

lib/licensed/sources/cocoapods.rb
```diff
@@ -0,0 +1,75 @@
...
+        @lockfile ||= Pod::Lockfile.from_file(config.pwd.join("Podfile.lock"))
+      end
+
+      def podfile
+        @podfile ||= Pod::Podfile.from_file(config.pwd.join("Podfile"))
+      end
+
+      def dependency_metadata(pod)
+        metadata = JSON.parse(Licensed::Shell.execute("pod", "spec", "cat", "--regex", "^#{pod.root_name}$"))
+        # The version returned by `pod spec cat` is the most recent version that exists which may not be the one installed.
```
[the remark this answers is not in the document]
LouisBoudreau: As decided [here](https://github.com/github/licensed/pull/584#discussion_r1045970349), I will remove this part of the code for performance reasons so no need to worry 😄

LouisBoudreau:
Thanks for you review!
I just wanted to let you know that I applied the requested changes.

lib/licensed/sources/cocoapods.rb
```diff
@@ -0,0 +1,69 @@
...
+
+      def lockfile
+        @lockfile ||= Pod::Lockfile.from_file(config.pwd.join("Podfile.lock"))
+      end
+
+      def podfile
+        @podfile ||= Pod::Podfile.from_file(config.pwd.join("Podfile"))
+      end
+
+      def dependency_metadata(pod)
```
jonabc, line 56: this function is no longer used, can you remove it?

jonabc (commented):
one last requested change to remove the method for gathering dependency metadata, otherwise this looks great! Once that's done and CI is green 🚀 :shipit:

jonabc:
> Thanks for you review!
> I just wanted to let you know that I applied the requested changes.

I appreciate the extra notice, I had another pretty busy week and then took a few days away from the computer to relax.  I should be more available to look at any more changes to your open PRs over the next few days.

LouisBoudreau:
No problem, I just applied the last requested change 😄

jonabc:
I haven't gotten around to fixing the cached license verification action to work with forks.  Everything else is looking good, merging!  Thanks again for putting this together.  :shipit:

jonabc (approved):

PR #592, opened 2022-12-22
Try out built-in caching for setup-python action

jonabc:
This _should_ work for the pip tests, but may not work for pipenv tests because the `Pipfile.lock` fixture doesn't get checked in...

Edit: the cache update didn't work for pipenv so I reverted that.  I'm adding some more caching in the workflow where it's built into setup actions in this PR, as well as some tidiness updates in general.


jonabc:
The `licensed-ci` GitHub Action has created a [pull request](https://github.com/github/licensed/pull/593) containing license metadata updates based on the changes in this branch.

PR #594, opened 2022-12-23
Move gitignore and licensed configurations for test fixtures into fixtures folders

jonabc:
A little more hygiene work, this change moves to make each test fixture folder more self-contained.  There are many files changed in this PR, but generally each file only has a small change and the changes follow common patterns.  It may be easier to review this as individual commits rather than all files in the PR at the same time.

1. Moved the fixture-specific gitignore rules from the root `.gitignore` into separate files in each fixture folder
1. Moved the fixture-specific licensed configurations from `test/fixtures/command` into each fixture folder and updated the paths as appropriate
1. Updated each source-setup script's force options to make better use of associated source CLI tooling if available, and prefer to use `git clean -ffX .` over `find ... | xargs ... rm -rf` to make git the source of truth about which files should be preserved


jonabc:
The `licensed-ci` GitHub Action has created a [pull request](https://github.com/github/licensed/pull/595) containing license metadata updates based on the changes in this branch.

PR #602, opened 2023-01-02
Remove support for old versions of npm and go

jonabc:
This PR will remove all support for older versions of npm and go, where the older versions are now end-of-life'd.  Generally, licensed tries to maintain support for actively supported versions of tools to keep the code in the repo easier to maintain.  With the upcoming major version bump in licensed to v4.0.0, it is a good time to cleanup some logic specific to older versions of tools which could be a breaking change to some users.

Go supports the most recent two minor release versions, [which are 1.18 and 1.19](https://endoflife.date/go).  Licensed supports one additional version beyond that, 1.17.  The go source had some logic related to using the `go` CLI at versions `< 1.11.0` which has been removed.

~~Node's oldest LTS major version is [14](https://endoflife.date/nodejs), where node 14.0.0 corresponds to npm version 7.10.  The npm source included a version check where the CLI args need to be different for versions `>= 7.0.0`.  I've made that behavior the default now with the removal of node 12 support.  In this case I've also updated the tested versions of the npm source, removing 12 and adding 18.~~

EDIT:  I don't know what I was looking at, node 14.0.0 corresponds to node 6.14.4.  I've reverted the changes to the npm source and tests, and only updated the tested versions in CI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -252,7 +252,7 @@ jobs:
     needs: core
     strategy:
       matrix:
-        node_version: [ 12, 14, 16 ]
+        node_version: [ 14, 16, 18 ]
     steps:
     - uses: actions/checkout@v3
     - name: Setup node
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -252,7 +252,7 @@ jobs:
     needs: core
     strategy:
       matrix:
-        node_version: [ 14, 16, 18 ]
+        node_version: [ 12, 14, 16 ]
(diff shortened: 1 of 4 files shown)


jonabc:
well that's odd.  After the CI failure I took another look at the npm version that ships with each version of node and it looks like the bump to 7.10 happens with node 16, where node 14 still uses the 6.x npm major version 🤦 .  I'm not sure how I missed that the first time around.  Anyway, I've reverted that commit and only updated the versions tested in CI to remove node 12 and add node 18

PR #608, opened 2023-01-08
build: remove GO111MODULE env as it is default on

chenrui333:
- dep has been archived in favor of go modules since early 2020, https://github.com/golang/dep/pull/2233
- `go modules` now is the default mode, so no need to set `GO111MODULE`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -182,8 +182,6 @@ jobs:
       run: script/source-setup/go
     - name: Run tests
       run: script/test go
-      env:
-        GO111MODULE: "on"
 
   gradle:
     runs-on: ubuntu-latest


chenrui333:
> 😢 removing the dep source would be a breaking change which I would prefer not to happen for a little while given the 4.0 release just went out with breaking changes. I'm trying to limit the amount of significant churn on the project and keeping dep support shouldn't hurt or interfere with licensed usage.
> 
> Would you mind reverting the changes related to the dep source and limit this PR to the `GO111MODULE` cleanup? Please create an issue to remove the dep source, and it can be handled as part of the next major version release with other breaking changes.

Yeah, that makes sense.

chenrui333:
@jonabc should be good now. :)

jonabc:
the bundler 2.4 failure is known and was fixed by a separate PR I merged in the last hour or so.  👍 merging and thanks for the update

PR #610, opened 2023-01-10
Fix gradlew location when trying to location a gradle executable

LouisBoudreau:
The https://github.com/github/licensed/pull/606 fix changed the search location of the `gradlew` executable
According to the [documentation](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper) the `gradlew` should be present at the root of the project  :
> A Gradle project typically provides a settings.gradle file and one build.gradle file for each subproject. The Wrapper files live alongside in the gradle directory and the root directory of the project. 

This PR therefore aims at updating where licensed searches for `gradlew`

### DIFF
--- a/lib/licensed/sources/gradle.rb
+++ b/lib/licensed/sources/gradle.rb
@@ -66,7 +66,7 @@ module Licensed
         return @executable if defined?(@executable)
 
         @executable = begin
-          gradlew = File.join(config.pwd, "gradlew")
+          gradlew = File.join(config.root, "gradlew")
           return gradlew if File.executable?(gradlew)
 
           "gradle" if Licensed::Shell.tool_available?("gradle")


lib/licensed/sources/gradle.rb
```diff
@@ -66,7 +66,7 @@ def executable
         return @executable if defined?(@executable)
 
         @executable = begin
-          gradlew = File.join(config.pwd, "gradlew")
```
LouisBoudreau, line 69: For gradle multi-projects builds, the `config.pwd` is often a subfolder and not the root of the project.
I think licensed should look for gradlew at the root since it is the official location for the gradle wrapper

What do you think about that ?

jonabc:
> The https://github.com/github/licensed/pull/606 fix changed the search location of the gradlew executable

Could you explain this a bit more?  The `config.pwd` value used when finding the executable is the same value that was passed as the root path into `Gradle::Runner.new`.  I don't see why/how the path would have changed with that PR.

Otherwise, I want to caution any expectation that `config.root` is equivalent to `project root`.  If not explicitly configured, it's most likely going to be the git repository root.  For monorepos which have multiple languages or projects within the same repo, it's common to have a single licensed configuration at the root of the repo with many `apps` for each of the many projects in the monorepo.

What do you think about adding a `gradle.gradlew` config setting to path off of `config.root` rather than expecting that `config.root` points at any specific directory?  It could default to assuming that `config.root` is the same folder as `settings.gradle` if that's going to be the most common arrangement, which is what I think you've implemented here.  Then users can change configure where to find `gradlew` if `config.root` points at any other location.

Hopefully that makes sense!

LouisBoudreau:
>  I don't see why/how the path would have changed with that PR.

After re-reading the changes and the [command class source code](https://github.com/github/licensed/blob/master/lib/licensed/commands/command.rb#L96), I confirm that I effectively made a false assumption that the change broke the search location for `gradlew`. Im sorry for this. What happened is that somewhere during the for the local development of https://github.com/github/licensed/pull/583 I installed `gradle` and forgot to delete it to make the final tests for our multi-project build. `licensed` was therefore using the `gradle` executable instead of the `gradlew` from our project.

Your idea to add a new property is really good! I implemented it, let me know if there is anything else I can do 😄

docs/sources/gradle.md
```diff
@@ -21,4 +21,13 @@ To run `licensed` for specific projects in a [multi-build project](https://docs.
...
+```
+
+### Gradlew 
+
+The `gradle.gradlew` property is used to determine where the `gradlew` executable is. The default location the [configuration root](../configuration/configuration_root.md).
+
+```yml
+gradle:
+  - gradlew: path/from/root/to/gradle/gradlew
 ```
```
jonabc, line 33: ❤️ thanks for the docs update!!

jonabc (approved):
🚀

docs/sources/gradle.md
jonabc: oh 🤦 the nested `gradlew` property isn't an array and shouldn't have the `- ` prefix.  I'll fix the docs in the release PR

PR #590, opened 2022-12-22
Cleanup ruby usage in workflows and gemspec dependency versioning

jonabc:
This PR is meant to improve the general hygiene of this repo relating to actions workflows and gem requirements

1. Updates actions workflows to remove specific caching of gems installed by `script/bootstrap`, and instead use the built-in caching mechanism from `ruby/setup-ruby`
   - `script/boostrap` is no longer used in CI
1. Updates actions workflows to use the ruby version from `.ruby-version` instead of manually setting a version in every job'
1. Updates `.ruby-version` to 3.1.3, the latest stable release, and updates the tested versions of ruby and bundler in CI
   - Licensed no longer needs to target ruby 2.6 now that exes aren't packaged and included in releases
1. Bumps the gem's required ruby version to `>= 2.6` to remove the explicit dependency on bundler
   - bundler is included in ruby 2.6+ at versions that licensed supports
1. Sets specific dependency versions in the gemspec file which should trigger dependabot to keep dependencies up to date

PR #619, opened 2023-01-26
Use version ranges for better downstream compatibility

jonabc:
The gem was updated to use explicit versions as an attempt to protect against breaking changes from non-major version updates and to make version usage more explicit.  I was also hoping that specifying explicit versions would make working with dependabot a little more straightforward as I wasn't seeing version update PRs occurring 🤷 

This change reverts part of https://github.com/github/licensed/pull/590 by changing version requirements back from an explicit requirement to a requirement range.  In order to mitigate some of the concerns with finding breaking changes from dependencies, I'm also un-ignored Gemfile.lock.  That seems to be the recommended strategy for gems these days anyway.


jonabc:
The `licensed-ci` GitHub Action has created a [pull request](https://github.com/github/licensed/pull/620) containing license metadata updates based on the changes in this branch.

jeffwidman:
> The gem was updated to use explicit versions as an attempt to protect against breaking changes from non-major version updates and to make version usage more explicit. I was also hoping that specifying explicit versions would make working with dependabot a little more straightforward as I wasn't seeing version update PRs occurring

I haven't double-checked the source to confirm, but I suspect the following prevented Dependabot from proposing version updates:
1. Since you didn't have a `Gemfile.lock` committed, Dependabot thought this was a library.
2. [For libraries, the default `versioning-strategy` is `increase-if-necessary`](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy).
3. Since the version was pinned explicitly, Dependabot probably has a check that says "if a library explicitly pins, leave it alone to prevent creating noise for the maintainer".

So you _probably_ could have worked around this by specifying a non-default `versioning-strategy`, like `increase`.

I suspect you're not interested in pursing this further, but if you are, then I'm happy to help with any pointers either here or via the internal GitHub channels.

jeffwidman:
I think you're good for now. 👍

PR #624, opened 2023-02-05
Add configured license amendment content

jonabc:
This change allows users to specify file paths to custom license amendment content provided by a library owner.  Amendment content files paths are specified per dependency type and name, and the files are treated as additional licenses for the dependency.  See the added [documentation](https://github.com/github/licensed/blob/319b85ec024f8378972ce8ff82667dc2fbaa87b8/docs/configuration/license_amendments.md) for details on how to use the added `amendments` configuration option.

/cc @mlinksva I'm curious for your thoughts here - is this an appropriate change?
1. It seems like there is some amount of risk in allowing users to specify content that would be treated as legally binding
2. Should the amendment content be treated as a legal notice rather than a license?

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -67,6 +67,13 @@ reviewed:
   - classlist # public domain
   - octicons
 
+# Specify license amendments that have been obtained from a dependency's owner
+# which alter the terms of the dependency's license 
+amendments:
+  bundler:
+    bcrypt-ruby:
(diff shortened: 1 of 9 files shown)


jonabc:
Ignoring the cabal failures, those tests are having issues during their setup steps.  I'm merging this in and will put out a release once the cabal tests are fixed in a separate PR.

Issue #625, opened 2023-02-07
Add support for pnpm as a source


rwwagner90:
```
Caching dependency records for ember-app
  npm
  * 0 npm dependencies
/usr/local/Cellar/licensed/4.0.4/libexec/gems/licensed-4.0.4/lib/licensed/sources/npm.rb:162:in `dig': String does not have #dig method (TypeError)

        parent&.dig("_dependencies", name) || peer_dependency(parent, name)
                    ^^^^^^^^^^^^^^^^^^^^^
	from /usr/local/Cellar/licensed/4.0.4/libexec/gems/licensed-4.0.4/lib/licensed/sources/npm.rb:162:in `extract_version'
	from /usr/local/Cellar/licensed/4.0.4/libexec/gems/licensed-4.0.4/lib/licensed/sources/npm.rb:74:in `block in recursive_dependencies'
	from /usr/local/Cellar/licensed/4.0.4/libexec/gems/licensed-4.0.4/lib/licensed/sources/npm.rb:68:in `each'
	from /usr/local/Cellar/licensed/4.0.4/libexec/gems/licensed-4.0.4/lib/licensed/sources/npm.rb:68:in `recursive_dependencies'
	from /usr/local/Cellar/licensed/4.0.4/libexec/gems/licensed-4.0.4/lib/licensed/sources/npm.rb:77:in `block in recursive_dependencies'
	from /usr/local/Cellar/licensed/4.0.4/libexec/gems/licensed-4.0.4/lib/licensed/sources/npm.rb:68:in `each'
	from /usr/local/Cellar/licensed/4.0.4/libexec/gems/licensed-4.0.4/lib/licensed/sources/npm.rb:68:in `recursive_dependencies'
	from /usr/local/Cellar/licensed/4.0.4/libexec/gems/licensed-4.0.4/lib/licensed/sources/npm.rb:77:in `block in recursive_dependencies'
	from /usr/local/Cellar/licensed/4.0.4/libexec/gems/licensed-4.0.4/lib/licensed/sources/npm.rb:68:in `each'
	from /usr/local/Cellar/licensed/4.0.4/libexec/gems/licensed-4.0.4/lib/licensed/sources/npm.rb:68:in `recursive_dependencies'
	from /usr/local/Cellar/licensed/4.0.4/libexec/gems/licensed-4.0.4/lib/licensed/sources/npm.rb:77:in `block in recursive_dependencies'
[...]

PR #626, opened 2023-02-08
Add pnpm source for dependency enumeration

jonabc:
Closes https://github.com/github/licensed/issues/625

This adds in basic support for to use pnpm as a source to enumerate dependencies.  The source uses the experimental (as of this writing) `pnpm licenses list` CLI command to enumerate data.  The only quirk with this source is that as far as I can tell there's no way to find dependencies individually per workspace project in repositories using pnpm workspaces.  Otherwise the source is very straightforward.

This PR includes everything needed to develop and test the source
- the source itself
- tests and test fixtures
- a source bootstrap script
- an entry in the tests GitHub Action workflow
- basic documentation about the source


rwwagner90:
This is awesome, thank you @jonabc!

jonabc:
I'm ignoring the cabal test failures for this change. Those test failures are coming from an issue during the setup phase of the tests and are almost certainly unrelated to adding this new dependency source.

PR #628, opened 2023-02-12
deps: bump ruby-version to 3.2.1

chenrui333:


jonabc:
@chenrui333 :wave: is the ruby version bump in this PR just to keep it as up to date as possible?  or does the version bump bring something of value specific to the project like a new ruby feature, specific performance improvement, etc?

PR #629, opened 2023-02-20
Add support for configuring reviewed and ignored dependencies at specific versions

jonabc:
This builds off of the configuration changes made in https://github.com/github/licensed/pull/560 to bring in full support for configuring reviewed and ignored dependencies that include version identifiers.

Version identifiers are given by adding `@...` to the end of dependency names in the reviewed and ignored configuration lists.  Identifiers can have the following values:
1. `*` - explicitly match all versions
1. a value that can be parsed by `Gem::Requirement` provides support for version ranges - `(=|<|<=|>|>=|~>) 1.2.3`.  For non-Ruby devs, the `~>` is called the twiddle-waka and is a pessimistic matcher similar to node's [tilda range operator](https://github.com/npm/node-semver#tilde-ranges-123-12-1).
   - Multiple ranges can be specified as a single string separated by commas - `dependency@>= 1.0.0, < 2.0.0`
1. a value that can't be parsed by `Gem::Requirement` will be compared with an exact string match to a found dependency's version

This mostly affects the `pnpm` source at the moment, which requires reviewing and ignoring dependencies using `@...`.  For other sources, the change is expected to be backwards compatible and users should only expect changes when configuring dependencies at specific versions or specific version ranges.

### DIFF
--- a/docs/configuration/ignoring_dependencies.md
+++ b/docs/configuration/ignoring_dependencies.md
@@ -17,3 +17,16 @@ ignored:
   go:
     - github.com/me/my-repo/**/*
 ```
+
+## Ignoring dependencies at specific versions
+
+Ignore a dependency at specific versions by appending `@<version>` to the end of the dependency's name in an `ignore` list.  If a dependency is configured to be ignored at a specific version, licensed will not ignore non-matching versions of the dependency.
+
+The version value can be one of:
+
+1. `"*"` - match any version value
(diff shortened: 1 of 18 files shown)


jonabc:
The gradle failures are related to a `current` version bump to 8.0.1 and not directly attributed to anything done here.  I've put together a [potential fix](https://github.com/github/licensed/pull/630) for that separately.  I'm going to merge this as-is, and when the gradle fix is looking alright both changes can be released with a minor version bump to licensed.

PR #630, opened 2023-02-21
Update gradle build files and tested gradle versions in CI

jonabc:
current gradle in CI was recently changed to 8.x, and the update of 7.9 -> 8.0 looks to have been causing [issues](https://github.com/github/licensed/actions/runs/4227384864/jobs/7341822341) due to a new [deprecation](https://docs.gradle.org/current/userguide/upgrading_version_7.html#old_artifact_transforms_api) with the versions of the kotlin jvm and gradle plugin that were used in the test fixtures' `build.gradle` files.  I made a few changes that I'm hoping will fix this.

- add previous current, 7.6, to run in CI
- update 6.4 -> 6.9.3
- bump plugins in build.gradle files causing CI failures

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -193,7 +193,7 @@ jobs:
         # are resolved, update this matrix list with '16'.
         # possibly fixed by https://github.com/jk1/Gradle-License-Report/pull/166
         java: [ '11' ]
-        gradle: ['current', '6.4']
+        gradle: ['current', '7.6', '6.9.3']
     steps:
     - uses: actions/checkout@v3
     - name: Set up Ruby
--- a/lib/licensed/sources/gradle.rb
+++ b/lib/licensed/sources/gradle.rb
@@ -143,7 +143,7 @@ module Licensed
           args = [command]
           # The configuration cache is an incubating feature that can be activated manually.
           # The gradle plugin for licenses does not support it so we prevent it to run for gradle version supporting it.
-          args << "--no-configuration-cache" if gradle_version >= "6.6"
+          args << "--no-configuration-cache" if gradle_version >= Gem::Version.new("6.6")
           Licensed::Shell.execute(@executable, "-q", "--init-script", @init_script.path, *args)
         end
 
@@ -164,7 +164,7 @@ module Licensed
                     }
                   }
                   dependencies {
-                    classpath "com.github.jk1:gradle-license-report:#{gradle_version >= "7.0" ? "2.0" : "1.17"}"
+                    classpath "com.github.jk1:gradle-license-report:#{gradle_version >= Gem::Version.new("7.0") ? "2.0" : "1.17"}"
                   }
                 }
 
@@ -194,6 +194,14 @@ module Licensed
             )
           f.close
           f
+        end
+
+        # Returns the version of gradle used during execution
+        def gradle_version
+          @gradle_version ||= begin
+            version = Licensed::Shell.execute(@executable, "--version").scan(/Gradle [\d+]\.[\d+]/).last.split(" ").last
+            Gem::Version.new(version)
+          end
         end
       end
     end
--- a/lib/licensed/sources/gradle.rb
+++ b/lib/licensed/sources/gradle.rb
@@ -9,7 +9,7 @@ require "fileutils"
 module Licensed
   module Sources
     class Gradle < Source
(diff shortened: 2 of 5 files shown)


jonabc:
I needed to make deeper changes to the code and tests to get things fully working for Gradle 8.  I was having trouble getting tests to work as I was expecting them to, and the following commits were made to try and get things in a more predictable place.

1. https://github.com/github/licensed/pull/630/commits/807d6fb3b8eab4ab4b3f4183cc2f2bec2a432070 updated the test file to simplify app configurations and to properly set `fixtures` folders for `multi_project` scenarios
1. https://github.com/github/licensed/pull/630/commits/4797fc524d50f330ba0aade14591760dbd784a54 then built on the `fixtures` fix by no longer passing directory paths into the gradle runner class.  This matches conventions and expectations used by other sources and simplifies the solution a bit.
   - at this point I also noticed that determining the project path wasn't necessary when configuration source paths were pointing at each individual project in a multi-project Gradle setup rather than the root.  Calling the Gradle executable is expensive and removing the call to determine the current project path should speed up the source a bit
   
At this point I was able to fix the problems that surfaced after [updating plugin versions](https://github.com/github/licensed/pull/630/commits/7c00e00fca05d9b67c33a1c0d04f826f76ee0b77)
1. https://github.com/github/licensed/pull/630/commits/c5c5c2d45375fa25b26edc6af1f32ba5b529ee5b replaces the `runtime` configuration with `runtimeOnly` and then updates the init script to create custom configurations that extend from the users requested configurations.  Using the `runtimeOnly` configuration in the init script caused problems because the configuration has `canBeResolved = false` which broke everything.  The error message suggested creating new configurations that extend the existing ones to work around the issue, which was faster than trying to dig into the Gradle code itself and it's MANY layers of interfaces, abstract classes, delegate classes and all sorts of enterprise-y complication goodness.

Along the way I also found a bug fix and a small code cleanup
1. https://github.com/github/licensed/pull/630/commits/84731f3295763860e9915c4141cd67b214310366 fixed a correctness issue where string comparisons were used for version checks.  this is a bug waiting to happen when gradle reaches major version 10, and is fixed by performing version checks with `Gem::Version`
   ```
   irb(main):001:0> "10" > "7"
   => false
   irb(main):002:0> Gem::Version.new("10") > Gem::Version.new("7")
   => true
   ```
1. https://github.com/github/licensed/pull/630/commits/81be51a0cfaa1fc7deb20f35a33b095051e80ce1 reuses a constant to reduce the likelihood of getting path mismatches in future changes to this source, and to more clearly tie the code that creates license CSV output with the code reading the CSV.

jonabc:
@LouisBoudreau :wave: would you be able to take a look at this change, specifically https://github.com/github/licensed/commit/c5c5c2d45375fa25b26edc6af1f32ba5b529ee5b?  It looks like that change is working, but I'm not sure if it's the best solution.

> https://github.com/github/licensed/commit/c5c5c2d45375fa25b26edc6af1f32ba5b529ee5b replaces the `runtime` configuration with `runtimeOnly` and then updates the init script to create custom configurations that extend from the users requested configurations. Using the `runtimeOnly` configuration in the init script caused problems because the configuration has `canBeResolved = false` which broke everything. The error message suggested creating new configurations that extend the existing ones to work around the issue, which was faster than trying to dig into the Gradle code itself and it's MANY layers of interfaces, abstract classes, delegate classes and all sorts of enterprise-y complication goodness.

LouisBoudreau:
Yes, I will check this out friday!

LouisBoudreau (approved):
I made a really simple proposition [here](https://github.com/github/licensed/pull/631) that could be merged into this PR before merging this one.

Thanks for cleaning up the code and updating the gradle source 🙇

jonabc:
> I made a really simple proposition [here](https://github.com/github/licensed/pull/631) that could be merged into this PR before merging this one.

To close the collaboration loop, I closed that PR and will not be bringing any of the changes proposed into this PR before merging.  It's a good proposal and I'm going to open an issue to re-evaluate the default Gradle configurations used for dependency enumeration, but that change likely needs to coincide with a major version bump since there's a chance that it could be a breaking change for users.  If conversation on the issue shows that it wouldn't be a breaking change then it's also really easy to remove `runtimeOnly` as a default configuration in the future 👍

Japottatweet (commented):
I hope this fixes it.

Issue #636, opened 2023-03-05
Create cocoapods plugin for enumerating dependencies for licensed


jonabc:
Cocoapods recently released a 1.12 update that includes a relaxation on the activesupport requirement from `< 7` to `< 8`.  The version update was needed for licensed to be able to integrate with cocoapods, however with that work done I'm concerned that the versioning is going to cause issues again in the future.

Instead of keeping the integration that exists with cocoapods, I think a better solution is to create a cocoapods plugin similar to https://github.com/segiddins/cocoapods-dependencies that can be used by licensed.  That plugin doesn't look like a good fit for licensed usage because for basic `pod dependencies` usage it seems to just print the podfile.lock entries without any additional data like targets or path to the root pod folder.

The plugin could expose a new CLI command that would be used by licensed, more or less pulling the code from Cocoapods#enumerate_dependency into a separate gem. The plugin would output the resulting name, path, and version data to the CLI in an easily parseable format like JSON or YAML or similar.

```mermaid
sequenceDiagram
    participant L as licensed
    participant P as Cocoapods plugin
    L ->> P: call plugin CLI, e.g. `pod dependencies list --targets ...`
    P ->> P: enumerate dependency information
    P ->> L: JSON (or similar) formatted list of dependencies
```

This approach offers a few benefits:
1. It decouples cocoapods dependency requirements from licensed's dependency requirements, resolving any concerns about this problem in the future.
1. It removes any concern about bloating licensed's installation size from the full cocoapods gem
1. It makes the functionality usable for other enumeration tools

cc @LouisBoudreau does this seem reasonable?

LouisBoudreau:
This really seems like a great idea!

When implementing the support of cocoapods into `licensed` the activesupport dependency requirements and the conflicts that could emerge is not something I thought about. 

Just like you, I think the the most important thing for licensed is it's stability. So investing time to make the cocoapods integration less prone to requirements failure is a great idea

jonabc:
I've created the plugin [here](https://github.com/jonabc/cocoapods-dependencies-list) but haven't yet published it.  I'll make sure the integration with licensed works well before releasing the plugin to rubygems

jonabc:
This is shipped in licensed 4.3.0

PR #644, opened 2023-03-20
Cococapods enumeration using cocoapods plugin

jonabc:
closes https://github.com/github/licensed/issues/636

This PR adds support for cocoapods without introducing a direct dependency on the cocoapods-core gem.  See the linked issue for full context, with a TL;DR that the additional nested dependency version requirements on common gems included from cocoapods-core made it difficult for downstream licensed users to install licensed alongside other usages of those same dependencies.

As an alternative approach to taking a direct dependency on the cocoapods-core gem, I created a `pod` CLI plugin at [jonabc/cocoapods-dependencies-list](https://github.com/jonabc/cocoapods-dependencies-list) which outputs dependency information as JSON.  The plugin was made pretty specifically for licensed but I suppose it's possible for the data output to be used by other tools as well.  I [searched rubygems](https://rubygems.org/search?query=cocoapods-depend) for a plugin that might already support licensed's usage but it looked like all existing plugins output graph visualizations of dependencies.


jonabc:
cc @LouisBoudreau FYI I'm going to merge this and put out a new release of licensed with cocoapods enabled.  If you have cocoapods projects that you were planning to use licensed for, could you give the new release a try when it's out and let m know how everything works out between licensed, the cocoapods plugin, or anything else related to this integration?

Issue #656, opened 2023-05-24
Make licensed status detect dependencies that have been removed


Mathbl:
Hello,

Would it be possible for licensed status to detect when a record exists for a dependency, but that dependency is not present anymore? 

Thanks!

jonabc:
@Mathbl :wave: typically we expect that users will run the `cache` command after dependencies on a project change to update the stored record files, including removing any outdated record files.  Would that solve your need, or am I misunderstanding the request?

Mathbl:
Indeed, cache will do the trick! Our use case is that our CI runs `licensed status` on push and will fail in case a dev forgot to run `licensed cache`, hence my request. 

But, it can also be the dev responsibility to think about running the cache command when adding/removing dependency. You can close the issue if you feel like there is not much value for `status` to detect dependency removal 🙂.

jonabc:
Ahhh I see what you mean now.  From a usage perspective, would you want to see the `status` command treat these stale files as errors or warnings?  Errors are meant to be blocking and will cause the `licensed status` command to exit with a failing code (1), whereas warnings are meant to be non blocking and will return a success exit code.

Mathbl:
Good question! In the case of a missing license/notice, it could cause legal issues, so throwing an error seems correct. 

For an extra license/notice from a previously removed dependency, I guess this is less problematic.

If I only take our use case into account, I would like the CI build (running `status`) to fail, because if it's just a warning and the build succeed, the dev will probably not notice the warning at all.

What do you think?

Mathbl:
Configuration option seems like a great idea. Thank you 😃

jonabc:
Opened https://github.com/github/licensed/pull/657 which adds the configuration option to ignore, warn, or error on "stale" metadata files from `licensed status`.

jonabc:
@Japottatweet :wave: you can [configure the option in the licensed configuration file](https://github.com/github/licensed/blob/main/docs/commands/status.md#detect-and-alert-on-stale-cached-metadata-files).  If the configuration option doesn't work please make sure you're using licensed v4.4.0 or greater 🙏

PR #657, opened 2023-05-26
Notify on stale cached records

jonabc:
closes https://github.com/github/licensed/issues/656

This change adds a new configuration option `stale_records_action` which is used by `licensed status` to determine if and how to handle stored dependency metadata records that do not match any currently used dependencies.

The configuration option supports the following values:

1. `'error'`: Treat stale cached records as errors.  Licensed will output errors for any stale metadata files and will cause `licensed status` to fail.
1. `'warn'`, `''`, or unset (default): Treat stale cached records as warnings.  Licensed will output warnings for any stale metadata files but will not cause `licensed status` to fail.
1. `'ignore'`, any other value: Ignore stale cached records.  Licensed will not output any notifications about stale metadata files.

The default behavior when the value is unset is `warn` so that this will not cause any unexpected `licensed status` failures on version upgrade.

### DIFF
--- a/docs/commands/status.md
+++ b/docs/commands/status.md
@@ -31,6 +31,21 @@ A dependency will fail the status checks if:
    - If `license: other` is specified and all of the `licenses` entries match an `allowed` license a failure will not be logged
    - A `reviewed` entry must reference a specific version of the depdency, e.g. `<name>@<version>`.  The version identifier must specify a specific dependency version, ranges are not allowed.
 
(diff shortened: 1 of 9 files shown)


jonabc:
Ignoring the bower failure, this is an issue with downloading the jquery package off bower and not an issue with the source itself.  I've recommended [removing that source entirely](https://github.com/github/licensed/issues/658) as most of it's usage has been replaced by npm, yarn, pnpm, deno, etc...

Issue #658, opened 2023-05-26
Remove bower support


jonabc:
Bower is a pretty old JS package manager that doesn't see any modern development.  It's been [failing in CI](https://github.com/github/licensed/actions/runs/4978996755/jobs/8910086853) for a little while due to certificate issues while installing dependencies.

When the next round of breaking changes are made to licensed, consider removing bower support alongside [removing dep](https://github.com/github/licensed/issues/612)

PR #690, opened 2024-01-25
use action provided GITHUB token for licensed workflow

elrayle:
NOTE: Bower is a known test failure.  It can be ignored.  See Issue #658.


dangoor (approved):
The change looks good! The bower test failure is unrelated. (I wonder if bower is no longer supported and that's why the certificate is expired?)

PR #703, opened 2024-03-18
Update Pipfile to make test compatible with Python version in CI

lumaxis:
`Pipfile.lock` contained an entry that limited a package version to Python versions <3.11 which broke installs on the latest stable Python version. Removed the constraint from `Pipfile` and now running the test against a small matrix of Python versions, most notably the ones that ship with current Ubuntu versions.


ajhenry (approved):
Looks good to me!

PR #737, opened 2024-08-13
Ensure that we don't have a vulnerability from cabal

dangoor:
This makes sure we don't get a homepage URL from cabal that's too
long and causes a performance issue leading to a denial of service.

Fixes https://github.com/github/licensed/security/code-scanning/1


elrayle (approved):
Thanks for taking care of security issues.

PR #747, opened 2024-10-03
Upgrade nokogiri to 1.16.5 to address CVE-2024-34459

dangoor:
This upgrade is not critical, because it addresses and xmllint issue
and nokogiri doesn't expose xmllint to us.

Fixes https://github.com/github/licensed/security/dependabot/11

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -138,7 +138,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


dangoor:
We can't do this without requiring Ruby 3.0, and I'm not sure that's a step we want to take.

dangoor:
@mlinksva That's good to know! Then I think it's worth taking that step here as well.

PR #815, opened 2025-02-28
Re-enable pipenv tests with Python 3.x

nozaq:
This PR re-enables pipenv tests with Python 3.x in the test matrix. They were previously failing because pyzmq, which was locked to an older version, caused an installation error.

This PR includes the following changes:

- Update indirect dependencies in Pipfile.lock, upgrading pyzmq so it’s compatible with Python 3.x.
- Explicitly add lazy-object-proxy to the Pipfile to ensure it isn’t accidentally removed in future updates. This dependency is required for the test cases to pass.
- Add [a missing assertion](https://github.com/licensee/licensed/pull/815/files#diff-4083ab7b8cb5a2c91fb662ede71f58789b5fbb56d5d170fb45ded310180dd7e3R65).


MikeMcQuaid (approved):
Thanks @nozaq!

PR #816, opened 2025-03-01
Re-enable NuGet tests for .NET LTS versions

nozaq:
This PR re-enables NuGet tests in the test matrix. Both .NET 3.x and 5.x are incompatible with Ubuntu 24.04 due to a libssl version mismatch (https://github.com/dotnet/core/issues/4749).

Because .NET 3.x and 5.x have reached the end of support, this PR updates the test matrix to use the current LTS version (8.x) and the next LTS version (10.x). These newer versions are compatible with Ubuntu 24.04.

Ref: https://dotnet.microsoft.com/en-us/download/dotnet

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -290,27 +290,26 @@ jobs:
     - name: Run tests
       run: script/test npm
 
-  # TODO: fix this and re-enable the tests.
-  # nuget:
-  #   runs-on: ubuntu-latest
-  #   needs: core
-  #   strategy:
-  #     matrix:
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):
Thanks again @nozaq!

MikeMcQuaid:
@nozaq Have given you maintain access to the repository 🎉. I'll try to open some issues for stuff but would appreciate having you around to be able to review my PRs. Thanks again for your work, you rock ❤️

nozaq:
@MikeMcQuaid Accepted! Sure, I'll review new PRs as they come in 👍🏼

PR #817, opened 2025-03-04
workflow: pin action versions and migrate archived actions

nozaq:
This PR includes several improvements to the GitHub Actions workflows:

- Replace actions that are no longer maintained with alternatives:
  - Migrate `github/licensed-ci` to `licensee/licensed-ci` 😜 
  - Migrate `gradle/gradle-build-action` to `gradle/actions/setup-gradle`.
  - Migrate from `actions-rs/toolchain` to using the pre-installed `rustup` command([reference](https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-rust#specifying-a-rust-version)).
- Pin all external actions to commit hashes, with human-readable version comments.
  - I pined all actions regardless of creators, please let me know if it's intentional to pin `actions/*` to tags instead of commit hashes. It's also reasonable to do so to reduce dependency management cost if we trust `actions` organization.

### DIFF
--- a/.github/workflows/licensed.yml
+++ b/.github/workflows/licensed.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - uses: ruby/setup-ruby@32110d4e311bd8996b2a82bf2a43b714ccc91777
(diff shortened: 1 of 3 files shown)


MikeMcQuaid (approved):
Wonderful, thanks @nozaq!

PR #818, opened 2025-03-04
Fix CVE-2025-27221

nozaq:
> [!NOTE]
> This PR requires a major release since `required_ruby_version` has been updated.

This PR updates `nokogiri` to version 1.18.3 to address [CVE-2025-27221](https://github.com/licensee/licensed/security/dependabot/22). 
As `nokogiri` v1.18.3 no longer supports Ruby 3.0.x, the minimum required Ruby version has been upgraded to 3.1.0. 

Additionally, Ruby 3.3 and 3.4 have been added to the CI test matrix.


nozaq:
Ruby 3.0 was removed from the CI test matrix alongside the required_ruby_version update. It appears that core (3.0) must also be removed from the required status checks for this PR to be merged.

MikeMcQuaid (approved):
Thanks @nozaq! Feel free to update the CI configuration. I don't feel strongly about whether this needs a major release or not; if we think it does it may be worth considering if there's any other functionality we want/need to remove before doing that.

nozaq:
@MikeMcQuaid Thank you! I've just updated the required status checks: I removed core (3.0) and added core (3.3) and core (3.4).

I also don't have a strong opinion on whether we should make this a major release, so simply including it in the next minor release should be sufficient.

> If we think it does, we might consider whether there's any other functionality we want or need to remove before we do that.

Regardless of this PR, I agree that summarizing what to include in the next major release is a great idea. Even if it won’t happen immediately, creating a milestone and linking issues would be a good place to start?

MikeMcQuaid:
> Regardless of this PR, I agree that summarizing what to include in the next major release is a great idea. Even if it won’t happen immediately, creating a milestone and linking issues would be a good place to start?

Works for me, great idea!

nozaq:
Just created a milestone and link some existing issues to it: https://github.com/licensee/licensed/milestone/1

Please feel free to add issues🙂

PR #856, opened 2025-05-05
Fix for pip packages

nadavshatz:
Handle packages with the separator pattern in their info

An example package for this is `json5`.

if you run `pip show json5 zipp` for instance you'll see that the `---` is present in the licenses list which breaks the `licensed cache` run for projects that have these pip packages.

This fix covers that by using a custom separator. it also separates the calls to `pip show` and parallelizes them since in large projects this can take >10m just to load since pip by default does them one by one.


MikeMcQuaid (approved):
Nice work, thanks!

taltcher:
@MikeMcQuaid - could you bump the gem's version so we can have this fix available?

MikeMcQuaid:
@taltcher Yes, I'm doing so now. I merged this 15 minutes ago. A little patience please 😀

taltcher:
@MikeMcQuaid of course 😄 thanks a lot for the quick handling!

MikeMcQuaid:
@taltcher @nadavshatz Released now in v5.0.4

PR #952, opened 2026-05-18
Support Licensee v10 and Parallel v2

Bo98:
Seen downstream that this is blocking licensee 10.x and parallel 2.x. Luckily there's no real breaking changes to be concerned about.


Bo98:
As an aside, Dependabot isn't working in this repo because the minimum specified Ruby version in the gemspec is 3.1 but the lockfile contains gem versions only compatible with Ruby 3.2+

MikeMcQuaid:
Thanks @Bo98!

PR #953, opened 2026-05-19
Fix assertions in status tests

Bo98:
Saw locally that these tests weren't working locally.

They were written as:

```rb
assert dependency_report.errors.any? do |e|
  ...
end
```

which means:

```rb
assert(dependency_report.errors.any?) do |e|
  ...
end
```

but it looks like the intention should have been:

```rb
assert(dependency_report.errors.any? do |e|
  ...
end)
```


Bo98:
Gemfile.lock needs regenerating for CI to pass, #952 will handle that

MikeMcQuaid:
Thanks again @Bo98!

